PR #22, opened 2016-03-06
Make it compatible with Parse push

frantic:
Looks like Parse Cloud, when sending push notifications via GCM, [is storing all information about it in the `data` field](react-native-push-notification). This PR makes `react-native-push-notification` compatible with Parse Push, i.e. it will fallback to `data.alert` for notification's body and will use application's label as default title.


dstik:
+1

zo0r:
Implemented: https://github.com/zo0r/react-native-push-notification/commit/bdd63fae921bb31dc0cbd1cddcb530b932a4502f

PR #25, opened 2016-03-09
Android background notifications

joshblour:
PushNotificationsIOS delivers notifications to your react-native app even when it's not in the foreground. 

This update enables similar behaviour for Android. Additionally, it doesn't display a notification in the drawer if the message is blank. The use case for this is background-fetch.


caledhwa:
+1 - Serious no brainer. Simple code change. Just do it.

PR #44, opened 2016-04-13
Data payload support

### LINKED ISSUE #41: Clarification on Android notifications
Does this package allow Android notifications to work in a similar manner to iOS? We're having trouble figuring out the correct implementation.

In our component:
```js
componentDidMount() {
  PushNotification.configure({
    onRegister: this.onPushNotificationRegister,
    onNotification: this.onPushNotification,
    senderID: 'foo'
  });
}
```

**App in foreground**
Both iOS and Android call `this.onPushNotification`.

**App in background**
* iOS natively displays a notification and when you tap it the app takes focus and `this.onPushNotification` is then called.
* Android gets the notification in the background and instantly calls `this.onPushNotification` but we want this to work like iOS. If we add `notification` to the GCM payload (see [payload](https://developers.google.com/cloud-messaging/concept-options#payload) docs) then Android natively displays a notification for us (great) but when clicking on it, nothing happens in the app (not great). Instead, should we not include a `notification` and call `PushNotification.localNotification()` which some data passed in?

**App not open**
* iOS natively displays a notification and when you tap it the app takes focus and `this.onPushNotification` is then called.
* I _think_ Android does nothing unless you include the `notification` payload from gcm? We tried adding a `click_action` that opens the app when you tap the notification but then `this.onPushNotification` is never called.

If we can figure this out, let's update the package docs.

lukefanning:
@zo0r Hi there, I have added support for the data string in a GCM payload. This resolves #41. I have tested it and it works perfectly for GCM payloads that contain both a "notification" and "data" object. Let me know if you'd prefer "else" on both of these "data" checks.

### DIFF
--- a/RNPushNotificationAndroid/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/RNPushNotificationAndroid/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -56,6 +56,15 @@ public class RNPushNotification extends ReactContextBaseJavaModule {
             constants.put("initialNotification", bundleString);
         }
 
+        String data = intent.getStringExtra("data");
+        if (data != null) {
+          bundle = new Bundle();
+          bundle.putString("data", data);
+          bundle.putBoolean("foreground", false);
+          String bundleString = convertJSON(bundle);
+          constants.put("initialNotification", bundleString);
+        }
+
         return constants;
     }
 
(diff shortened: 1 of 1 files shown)


pewh:
Hi @lukefanning. Does it work on `PushNotification.localNotification()`? I'm trying to create app with local push notification (without GCM) and need to trigger action when clicked. Thanks :)

lukefanning:
@pewh The changes I made have no effect on the local notifications produced by this package and we aren't using them for our project so I can't really assist you much. I had a brief look at the code and I can see that he supplies a PendingIntent to the local notification meaning it should open the application when clicked. The intent contains the "notification" bundle which would still be picked up by this PR since I didn't touch that handling.

zo0r:
Support added in: https://github.com/zo0r/react-native-push-notification/commit/add425635b299965a74defce0b090a9e7b6e2b5a

lukefanning:
@zo0r Sorry just thought I'd mention, I'm not sure that you've handled hybrid GCM notifications fully. ```onMessageReceived()``` is not called if the GCM payload contains a ```notification``` object

PR #52, opened 2016-04-25
Unregister fix

lukefanning:
@zo0r Unregister was not removing the event listeners correctly as bind() returns a new function everytime it is called.

p.s. I am relatively new to git, soz for other PRs


lukefanning:
I made a further commit to this. This allows multiple event listeners to work with the correct ```this```. This bug wasn't evident before because event listeners weren't being removed properly in ```unregister()```

zo0r:
https://github.com/zo0r/react-native-push-notification/commit/131b38f8c5d184f1b0cf996fc61725ddea0b749d

PR #63, opened 2016-05-07
Initial work on scheduled notifications on Android

cancan101:


zo0r:
@cancan101 thanks for submitting this PR. Do you know if the AlarmManager will restore the pending notifications if the Android OS is rebooted?

cancan101:
By default it does not.

You can see here how others deal with the issue: https://github.com/Neson/react-native-system-notification/blob/6839d8f6df7ded95303331f0277a22a2b29f0de7/android/src/main/java/io/neson/react-native/notification/SystemBootEventReceiver.java but that does require having some way of serializing the notifications

PR #83, opened 2016-05-30
Allow latest RC version of React Native as peerDependency

robmoorman:
Without the ability to solve peerDependencies with NPM 3 it's temporary impossible to install RC versions of libraries (React Native) without allowing them in the peerDependencies. Since React Native is releasing RC's quite often it would be nice to make sure the latest RC is supported as peerDependency.

Maybe another solution is to use a wildcard instead.


robmoorman:
Version 0.27.2 is shipped out already, you can close this one

PR #86, opened 2016-06-02
Android: Handle nested notification bundle

jedlau:
Setup: A backend is sending push notifications via GCM using XMPP. The
message uses the notification payload format, which is a JSON object
(see
https://developers.google.com/cloud-messaging/xmpp-server-ref#downstream-xmpp-messages-json).

Issue: On an Android device, the notification payload arrives as a
nested Bundle with key "notification", within the outer "notification"
Bundle that is passed to notifyNotification(). In its current form,
convertJSON() doesn't convert this nested Bundle into a string, and as
such, the notification payload is missing from the notification that
is passed to the onNotification() callback.

Fix: In convertJSON(), call itself recursively to convert nested
Bundles into JSON. Change the function's return type to JSONObject.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -52,8 +52,8 @@ public class RNPushNotification extends ReactContextBaseJavaModule {
         Bundle bundle = intent.getBundleExtra("notification");
         if ( bundle != null ) {
             bundle.putBoolean("foreground", false);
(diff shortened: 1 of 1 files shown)


cmcewen:
👍

npomfret:
Are you able to re-do this? the code has changed significantly and the merge looks difficult.

npomfret:

sebo:
@jedlau I fixed the conflicts and added another pull request.  do you have time to test it? https://github.com/zo0r/react-native-push-notification/pull/358

Thanks

Dallas62:
Duplicate with #358

PR #98, opened 2016-06-16
new features for android notifications

npomfret:
* fixed 'id' and 'number' field parsing - docs specified (correctly) that these fields are numeric but the code was parsing them as strings which would return null
* added log warnings for when user uses strings instead of ints for 'id' and 'number' fields
* surrounded message parsing with a try/catch block to prevent malformed messages from crashing the app
* added support for 'tag' field
* added support for 'group' field
* added support for 'playSound' field
* added support for custom vibration ('vibrate' and 'vibration' field)
* updated docs to reflect new functionality

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,6 +21,12 @@ android {
     }
 }
 
+buildscript {
+    dependencies {
+        classpath 'com.android.tools.build:gradle:2.1.2'
+    }
+}
+
 allprojects {
     repositories {
         mavenLocal()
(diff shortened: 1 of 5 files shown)


fermuch:
(I'm not a java developer, so I'm not sure if what I'm saying is a problem with my setup or with this PR)

After testing your PR, I get the following error:

```
* What went wrong:
A problem occurred configuring project ':app'.
> A problem occurred configuring project ':react-native-push-notification'.
   > failed to find Build Tools revision 23.0.3
```

Changing the version to `23.0.0` fixed the issue, and everything seems to work alright.

npomfret:
Thanks for testing.  Not sure what that means, but cheers for fixing

android/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "23.0.3"
```
fermuch: This is the change that broke my build. Reverting to 23.0.1 fixed the problem.

npomfret:
https://github.com/zo0r/react-native-push-notification/pull/102

PR #101, opened 2016-06-17
Include data payload when opening from background notification 

cmcewen:
When clicking on a notification, I was getting the app to open but `onNotification` wasn't being fired since the Bundle wasn't nested under `notification`.


cmcewen:
Just for reference I'm using [node-gcm](https://github.com/ToothlessGear/node-gcm) and sending messages like this:
```
let message = new gcm.Message({
  priority: 'high',
  notification: {
    'title': job.data.title,
    'body': job.data.message,
  },
  data: {
    'url': job.data.deeplink
  }
});
```

cmcewen:
bumping this - anything I can do here?

PR #102, opened 2016-06-20
new features for iOS and Android

npomfret:
iOS:
* added optional iOS only fields alertAction & category
* implemented 'playSound' for iOS and updated docs to reflect that this field works on both android and iOS
* implemented 'number' for iOS which sets the badge number, and updated docs to reflect that this field works on both android and iOS
* fixed iOS notifcation callback object to not have 'undefined' in the optional fields that have not been set
* added userInfo  to docs

Android:
* fixed 'id' and 'number' field parsing - docs specified (correctly) that these fields are numeric but the code was parsing them as strings which would return null
* added log warnings for when user uses strings instead of ints for 'id' and 'number' fields
* surrounded message parsing with a try/catch block to prevent malformed messages from crashing the app
* added support for 'tag' field
* added support for 'group' field
* added support for 'playSound' field
* added support for custom vibration ('vibrate' and 'vibration' field)
* updated docs to reflect new functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ public class MainApplication extends Application implements ReactApplication {
 }
 ```
 
-Add `onNewIntent` module (in `MainActivity.java`)
+Add `onNewIntent` (in `MainActivity.java`)
 
 ```java
 import android.content.Intent; // <--- Import Intent


npomfret:
Anyone able to look at this?  They're all non breaking changes. I've tested on some iOS and android devices and it seems fine so far.

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -30,15 +33,15 @@ public RNPushNotificationHelper(Application application, Context context) {
...
-      String packageName = mContext.getPackageName();
-      Intent launchIntent = mContext.getPackageManager().getLaunchIntentForPackage(packageName);
-      String className = launchIntent.getComponent().getClassName();
-      try {
-          return Class.forName(className);
-      } catch (ClassNotFoundException e) {
-          e.printStackTrace();
-          return null;
-      }
+        String packageName = mContext.getPackageName();
```
varungupta85: This file contains a lot of differences which are purely formatting changes or the code is shuffled making it hard to review the changes. Would it be possible for you to remove all the formatting and shuffled code kind of changes and only have changes which are actual changes? I agree that formatted code is much better to read but I think PR containing formatting changes should just be a separate PR with no functionality changes which can be merged harmlessly. Thanks!

varungupta85:
File `RNPushNotificationHelper.java` contains a lot of differences which are purely formatting changes or the code is shuffled making it hard to review the changes. Would it be possible for you to remove all the formatting and shuffled code kind of changes and only have changes which are actual changes? I agree that formatted code is much better to read but I think PR containing formatting changes should just be a separate PR with no functionality changes which can be merged harmlessly. Thanks!

npomfret:
Sure.  How about the other way around - fist I send you a PR with ONLY formatting changes.  Then after another with the material changes.  I just think that would be easier for me.

varungupta85:
@npomfret I just wanted to clarify that I don't have write access to this repo. So, even if I review the changes, I won't be able to merge them. I am actually in the process of adding a few features to this package myself and I came across your PR and thought that I would just check your changes so that I don't make any breaking changes unknowingly. I am specifically adding features to be able to cancel a particular scheduled notification and to be able to play a custom sound when a notification is received instead of the default sound. The second feature may have some conflicting changes with your PR, so I wanted to maybe just get your changes as well regarding playing a sound. Anyhow, that's when I started to review your changes and left my earlier comment. It may still be easier for the actual reviewers if you make those changes. I can definitely leave a note after I review and even test a few of the features in Android. Thanks!

npomfret:
Oh I see, thanks for clarifying.  Maybe I'll wait until I hear from a contributor before making any further changes.

PR #119, opened 2016-06-29
Updated to use static method instead of this when popping initial notification

bennyguitar:
Was running into a problem where on Android, calling `this._onNotification` in this place resulted in an undefined `this._onNotification`. Switching to `Notifications._onNotification` seems to have fixed the problem.


npomfret:
I don't fully understand, the `_onNotification` is static and as far as I can see shouldn't be called statically.  Can you check if you are still getting the problem in the latest release?

npomfret:

PR #123, opened 2016-07-01
Add ability to push silently without notification

jounistam:
Allows service to send data without showing any notification to user when app is not running.


jeveloper:
Hey @jounistam , this is already available. When adding notification handler, data of a message can be without Title or subject (this way ios and android ) and it won't be displayed to the user, but your app can use data passed. See google's format (mainly free json) but apple uses an array.

npomfret:
Can I close this if it is already available?

jeveloper:
Sure @npomfret , i believe i was able to explain how this is can be done :)

Issue #122, opened 2016-07-01
Capturing clicks on Android


iggyfisk:
Foreground/background notifications are inconsistent at the moment (a common theme for all multi-platform push notification libraries).

On iOS, receiving a remote notification in foreground immediately triggers the `onNotification` event handler with `foreground: true`. If the app is in background when the notification arrives, clicking the notification will trigger the `onNotification` handler with `foreground: false`.

On Android, foreground notifications work the same. But in background, when the notification arrives, the `onNotification` handler triggers **immediately** with `foreground: false`. Clicking the notificiation opens the app of course, but doesn't trigger the event handler again. 

My goal is to execute some code if **and only if** the user clicks on a certain notification. Are there any plans to homogenize the behavior across platforms, or is there another way to determine when notifications are clicked on Android?

zo0r:
Try `userInteraction` in **2.0.1**

iggyfisk:
This is nice and consistent on both platforms now, receiving in foreground gives `foreground: true`/`userInteraction: false`. Clicking a notification `foreground: false`/`userInteraction: true`. Thanks!

iggyfisk:
I forgot to say, in my original issue I wrote:
> Clicking the notificiation opens the app of course, but doesn't trigger the event handler again. 

This was a bug in my setup, because our app as two activities (one for a first-time splash screen, and a different MainActivity) so `onNewIntent` in MainActivity.java was never called. Clicking the notification meant that `onCreate` in our SplashActivity.java was called instead. Thought I'd mention it it in case someone googles their way in here.

osungjin:
@iggyfisk How can receive trigger if use rn 0.30+ ? It same case as two activities.

Dean177:
@iggyfisk Having a similar issue due to implementing a seperate splash screen. How did you work around this?

iggyfisk:
@osungjin @Dean177 it seems there's a clean, permanent solution on the way https://github.com/zo0r/react-native-push-notification/issues/183, I came up with a hack until then.

After `onNewIntent` went away in RN 0.30, my workaround is adding the INFO intent to our MainActivity in AndroidManifest.xml. Like so:

```XML
<activity
  android:name=".MainActivity"
  android:label="@string/app_name">
  <intent-filter>
      <action android:name="android.intent.action.MAIN" />
      <category android:name="android.intent.category.INFO" />
  </intent-filter>
</activity>
<activity
  android:name=".SplashActivity"
[...]

PR #125, opened 2016-07-05
Make notification push behave similar to iOS

whs:
This PR disable intent broadcast in Android when the application is in the background and the push message will generate a notification.

This make the Android's behavior similar to iOS: Only one push message will be received: when it is received by active application, or when the notification is tapped.

Potential breakage when the application rely on the background message to do something, but such behavior won't work in iOS anyway.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -39,13 +39,14 @@ public class RNPushNotificationListenerService extends GcmListenerService {
     private void sendNotification(Bundle bundle) {
 
         Boolean isRunning = isApplicationRunning();
-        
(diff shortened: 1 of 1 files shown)


varungupta85:
In iOS application, I have to implement background-fetch to run some code in the background when a push notification is received. Here, I think it was done due to the broadcasted intent. Does it mean that with your changes, the application will not wake up to perform some work when a push notification is received?

whs:
Yes, this will prevent the application from waking up. However, if the push data does not contains the key `message` the old behavior is retained. You'll have to use other way to generate the notification (I believe local notification works? or you can use [GCM's notification key](https://developers.google.com/cloud-messaging/http#message-with-payload--notification-message))

I couldn't find any reliable way to detect whether the notification is received in background, or from notification tap (which are also mentioned in unanswered issues #41 #122) so I think this is the best way to implement it.

whs:
I'm retracting this PR. Use `userInteraction` in the latest version instead.

Issue #72, opened 2016-05-16
Android: 'onNotification' not executing when app is closed and clicked to notification


rkostrab:
[android] 
Hello, when I put app in the background and I get push notification then `onNotification` is executed as I expect. But when I close app (using android back button) and I get push notification then `onNotification` is not executed.

I know `onNotification` is fired from Java in `onNewIntent` method. But I really need to pass some data from notification into JavaScript when application was closed before and I click on notification.

How to do it?

j0hn:
I'm having this same issue. I'm using [node-gcm](https://github.com/ToothlessGear/node-gcm) to send the notification to the app.

eduardb:
Same here. Some help on how to overcome this issue would be great.

eduardb:
UPDATE: It seems that onNotification is called after all when opening the app from the notification, but for some reason the YellowBox was not appearing when calling `console.warn()`. `Alert.alert()` does the trick though.

j0hn:
@eduardb even when your application is completely closed?

eduardb:
Completely, as in swiped away from the recents screen, yes.

j0hn:
Awesome, i couldn't achieve that. Would you be so kind to share some
example on how to make it work?

eduardb:
Have a look over this PR: https://github.com/novoda/spikes/pull/57

florindobre:
@eduardb how exactly do you have the Alert firing? I have an alert on the first line in onNotification, and it's working fine when the app is closed for iOS, but for Android it's not triggering.

florindobre:
After some debugging it looks like the `onMessageReceived` method from the `RNPushNotificationListenerService` class gets called when receiving the notification while the app is closed/in the background. But clicking on the notification itself only launches the app and does not pass its data to React.

@zo0r : any ideas on this? iOS working fine with clicking on notifications.

rskull:
if revert this commit, It work?

https://github.com/zo0r/react-native-push-notification/commit/1aa5c9684d8481428860d866ec50dea968cb19b8

npomfret:
I don't know if its related but I never get `onNotification`, weather the app is in the foreground or background.

https://github.com/zo0r/react-native-push-notification/issues/93

rskull:
@npomfret https://github.com/rskull/react-native-push-notification
I was solved with this repository.

florindobre:
[...]

PR #127, opened 2016-07-06
Add missing this binding to popInitialNotification callback

levito:
When popInitialNotification is true and app is closed, clicking on a notification causes an app crash


levito:
This should fix #72

PR #133, opened 2016-07-07
Defined suffix-less index.js

jevakallio:
When running unit tests against a codebase that imports `react-native-push-notification`, the test fails because it's not able to import `./component`.

Create a dummy file to fix broken import. Test author can mock the module definition as they please, but the default behaviour should not prevent other tests from running, as it does now.


yonahforst:
please merge :)

PR #148, opened 2016-07-20
just formatting - NO funcational change

npomfret:
In order to make future changes easier I've just hit the reformat key on the java code in Android Studio.  There are NO functional code chages here


kfiroo:
@npomfret @zo0r looks good

PR #152, opened 2016-07-25
Use onNewIntent from ActivityEventListener

Bhullnatik:
Hello,

React Native a new event in the 0.30 version , [`onNewIntent`](https://github.com/facebook/react-native/commit/2fc0f4041ec836c1e1aef1f5d9c77b4f4aba4aae), which allow us to know when the current activity receives a new intent without having to implement a method in `MainApplication` and `MainActivity`.

It simplifies the code a lot, even when implementing the module since we don't need to pass the event from the activity -> application -> module.

I tested it, it works fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ In your `AndroidManifest.xml`
 				<category android:name="${applicationId}" />
 			</intent-filter>
 		</receiver>
-	
+
 		<receiver android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationPublisher" />
 		<service android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationRegistrationService"/>
 		<service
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -90,53 +90,23 @@ import com.dieam.reactnativepushnotification.ReactNativePushNotificationPackage;
 
 public class MainApplication extends Application implements ReactApplication {
 
-  private ReactNativePushNotificationPackage mReactNativePushNotificationPackage; // <------ Add Package Variable
```
Bhullnatik: Since there are no more irregular steps to set up the package, maybe we should mention [rnpm](https://github.com/rnpm/rnpm) instead of having a step-by-step guide?

zo0r:
@Bhullnatik can you sync with master and send the **PR** again?

Bhullnatik:
@zo0r Done! Sorry for the delay.

zo0r:
@Bhullnatik thanks! 👍

PR #154, opened 2016-07-25
bug fix to some parsing

npomfret:
Code to parse some numberic fields was rejecting valid numeric data, and only accepting invalid string types


npomfret:
@zo0r I made a small mistake in my previous PR, aplogoies.  This should fix: https://github.com/zo0r/react-native-push-notification/pull/154

PR #168, opened 2016-08-18
Android - added ShortcutBadger support

kfiroo:
A notification object containing `badge` field will set the the `unread notification counter` on the app icon.

A notification without the `badge` field will not affect the counter.

To reset the counter set the `badge` field to 0.

See ShortcutBadger for supported launchers.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -31,5 +31,5 @@ dependencies {
     compile 'com.android.support:appcompat-v7:23.1.1'
     compile 'com.facebook.react:react-native:+'
     compile 'com.google.android.gms:play-services-gcm:7.8.0'
-    compile 'me.leolin:ShortcutBadger:1.1.6@aar'
+    compile 'me.leolin:ShortcutBadger:1.1.4@aar'
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,9 @@
 apply plugin: 'com.android.library'
 
+repositories {
+    mavenCentral()
+}
+
 android {
     compileSdkVersion 23
     buildToolsVersion "23.0.1"
@@ -27,4 +31,5 @@ dependencies {
     compile 'com.android.support:appcompat-v7:23.1.1'
     compile 'com.facebook.react:react-native:+'
     compile 'com.google.android.gms:play-services-gcm:7.8.0'
+    compile 'me.leolin:ShortcutBadger:1.1.6@aar'
 }
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -231,6 +231,21 @@ public class RNPushNotificationHelper {
                 notification.setVibrate(new long[]{0, vibration});
             }
 
+            int badge = -1;
+            if (bundle.containsKey("badge")) {
+                try {
+                    String badgeAsString = bundle.getString("badge");
+                    if (badgeAsString != null) {
+                        badge = Integer.parseInt(badgeAsString);
+                    }
+                } catch (Exception e) {
+                    badge = -1;
+                }
+            }
+            if (badge > -1) {
+                ShortcutBadger.applyCount(mContext, badge);
+            }
+
             Notification info = notification.build();
             info.defaults |= Notification.DEFAULT_LIGHTS;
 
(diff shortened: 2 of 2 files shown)


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -231,6 +232,21 @@ public void sendNotification(Bundle bundle) {
                 notification.setVibrate(new long[]{0, vibration});
             }
 
+            int badge = -1;
+            if (bundle.containsKey("badge")) {
```
npomfret: Can I suggest you use the 'number' field so its the same as iOS?

kfiroo: According to apple `badge` is the right field to use
https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/TheNotificationPayload.html

npomfret: Actually, after a read of the RN [docs](https://facebook.github.io/react-native/docs/pushnotificationios.html) I think it's applicationIconBadgeNumber when included in a notification. 

And there's also a static method `setApplicationIconBadgeNumber`.

npomfret: ...however in this project, if you take a look at [index.js](https://github.com/zo0r/react-native-push-notification/blob/master/index.js#L121) you can see that the `number` field seems to get translated into `applicationIconBadgeNumber` for iOS.

igrayson: If we want iOS parity, `badge` is right key to read. The doc linked is pretty clear, and we can confirm. We use `badge` payload attribute (along side `alert`, `content-available`, and `sound` -- also discussed in the doc)  for our existing iOS-only app.

android/build.gradle
```diff
@@ -27,4 +31,5 @@ dependencies {
     compile 'com.android.support:appcompat-v7:23.1.1'
     compile 'com.facebook.react:react-native:+'
     compile 'com.google.android.gms:play-services-gcm:7.8.0'
+    compile 'me.leolin:ShortcutBadger:1.1.4@aar'
```
igrayson: Suggest using the latest `1.1.8`

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -231,6 +232,21 @@ public void sendNotification(Bundle bundle) {
                 notification.setVibrate(new long[]{0, vibration});
             }
 
+            int badge = -1;
```
igrayson: More clear: `Integer badge;`

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -231,6 +232,21 @@ public void sendNotification(Bundle bundle) {
...
+                try {
+                    String badgeAsString = bundle.getString("badge");
+                    if (badgeAsString != null) {
+                        badge = Integer.parseInt(badgeAsString);
+                    }
+                } catch (Exception e) {
+                    badge = -1;
+                }
+            }
+            if (badge > -1) {
```
igrayson: using `Integer`:
```java
if (badge != null && badge >= 0) {

}
```

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -231,6 +232,21 @@ public void sendNotification(Bundle bundle) {
                 notification.setVibrate(new long[]{0, vibration});
             }
 
+            int badge = -1;
+            if (bundle.containsKey("badge")) {
+                try {
+                    String badgeAsString = bundle.getString("badge");
```
igrayson: This seems wrong; the payload should contain `badge` as a numeric, not a string.

igrayson:
I don't think `RNPushNotificationHelper` is the right place to put this logic. `RNPushNotificationHelper` is not invoked for incoming remote push notifications, while the app is in the foreground.

This logic probably belongs in `RNPushNotificationListenerService`

igrayson:
I've pushed a PR with the approach we've been using: https://github.com/zo0r/react-native-push-notification/pull/181

npomfret:
So will `applicationIconBadgeNumber` work for both iOS and Android?  And will the `number` field remain in use for local notifications, because that is whats being used to create _ applicationIconBadgeNumber_ right now.

PR #193, opened 2016-09-07
Add Activity argument to onActivityResult

idris:
This changed in react-native v0.33.0-rc.0

See https://github.com/facebook/react-native/commit/fbd2e139103e3d520f6dfc009d6200f8b8168e35


idris:
@zo0r can you release a new version to npm?

Also, maybe we should change the peerDependency to `"react-native": ">=0.33"` since this probably won't play nicely with previous versions

Issue #219, opened 2016-09-30
Can't receive notifications when app is closed or in background


ddobby94:
manifest file:

```
<uses-permission android:name="android.permission.WAKE_LOCK" />
<permission
        android:name="co.my.packagename.permission.C2D_MESSAGE"
        android:protectionLevel="signature" />
<uses-permission android:name="co.my.packagename.C2D_MESSAGE" />
<uses-permission android:name="com.google.android.c2dm.permission.RECEIVE" />
<uses-permission android:name="android.permission.VIBRATE" />
<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
```

in the application:
         `<meta-data
                android:name="com.google.android.gms.version"
                android:value="@integer/google_play_services_version" />`

```
    `<receiver
            android:name="com.google.android.gms.gcm.GcmReceiver"
            android:exported="true"
            android:permission="com.google.android.c2dm.permission.SEND" >
        <intent-filter>
            <action android:name="com.google.android.c2dm.intent.RECEIVE" />
            <category android:name="co.my.packagename" />
        </intent-filter>
    </receiver>`

    <receiver android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationPublisher" />
    <receiver android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationBootEventReceiver">
        <intent-filter>
            <action android:name="android.intent.action.BOOT_COMPLETED" />
        </intent-filter>
    </receiver>


    <service android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationRegistrationService"/>

    <service
            android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService"
            android:exported="false" >
        <intent-filter>
            <action android:name="com.google.android.c2dm.intent.RECEIVE" />
        </intent-filter>
    </service>
```

settings.gradle:

`include ':app'
...
include ':react-native-push-notification'
project(':react-native-push-notification').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-push-notification/android')`

build.gradle:

```
dependencies {
    classpath 'com.android.tools.build:gradle:1.2.0'

}
```

`

app/build.gradle
`
dependencies {
    compile fileTree(dir: "libs", include: ["*.jar"])
    compile "com.android.support:appcompat-v7:23.0.1"
    compile 'com.android.support:support-v4:19.1.+'
   ...
    compile project(':react-native-push-notification')

```
[...]

Issue #217, opened 2016-09-29
How get notification event when app is killed or in background


jitenderchand1:
When my app is killed or in background and notification arrives then I want to set some key in storage. So that when user open my app I can show different list.

onNotification event is not working when app is killed or in background

ddobby94:
+1

abdurrahmanekr:
Hi @jitenderchand1 **RNPushNotificationListenerService > onMessageReceived** its worked ! [this commit](https://github.com/Detaysoft/react-native-push-notification/commit/7cf94d441c93c6b7b4faa4692575572b911084be) simple sqlite save

Issue #170, opened 2016-08-23
onNotification does not fire when received on Android


JonesN3:
Is this feature not supported on android, or am I doing something wrong?

heminm:
What version of the package and RN are you using? Are you trying local or remote notification?

JonesN7:
RN 30, local notifications.

vikeri:
I have the same issue with local notifications, when I schedule one and it fires `onNotification` is not triggered. But if I click the notification it is.

juanpasolano:
Same issue:
RN 0.30, 
react-native-push-notification 2.0.2

barroslee:
I got same issue with local notifications and RN 0.33.0. `onNotification()` is not called until I click the notification. I just add `notifyNotification()` in `sendNotification()` for a temporarily solution....

npomfret:
So the previous behaviour was for onNotification() to be called immediately after sending a local (or scheduled) notification?

Does this only affect Android?

Has it started happening since RN 0.33?

vikeri:
It never worked for me. I added this library around 0.31.0 only doing Android for now though. Looked through the Java code but could not find any callback to JS in the code that runs when the notification is fired.

vikeri:
i.e. I'm not sure it ever worked as @npomfret describes

npomfret:
Sorry, I was just asking a question.  In order to fix it I need to know what is the desired / expected behaviour for receipt of a local notification?  What does the iOS implementation do?

barroslee:
Hi @npomfret , I am not sure what's the real behavior. Just because I click the notification, `onNewIntent()` will be called. then `notifyNotification()` will fire event 'remoteNotificationReceived' to trigger onNotification. So I add it in `sendNotification()`.

In ios when app is foreground, the notification never appear but just trigger onNotification.

I upgrade to RN 0.33 from RN 0.30. Both of them also have this issue.

npomfret:
I think, but am not sure, the iOS behaviour is that local notifications just go into the notification center and no events are fired.  When the user clicks on a notification it is then that the _onNotification_ js call happens.  Is the Android behaviour is the same right now?  If so, I would suggest there isn't a bug here.

I don't quite understand how the `remoteNotificationReceived` triggers _onNotification_?  Looking at the docs for iOS and js code it seems that the only events being listened to are register, notification, localNotification and remoteFetch.  Have you any idea how this works?

npomfret:
[...]

Issue #161, opened 2016-08-03
Callback not triggered when notification is opened and app is killed


andy9775:
I'm having an issue with the `onNotification` callback not being called if the app is killed and the notification is clicked. Is there anyway to implement this functionality in the app?

RN -> 0.30.0
react-native-push-notification -> 2.0.2

parrunset:
same here

jakkra:
Ok, so I managed to work around the issue. However, I'm running a version behind you guys, but it might still work for you.
RN -> 0.29.0
react-native-push-notification -> 2.0.1

```
PushNotification.configure({ popInitialNotification: false, ...})
```

And in my onRegister method, I did:
```
PushNotification.popInitialNotification((notification) => {
      if (notification) { this.handleNotification(notification); }
});
```

kfiroo:
@jakkra note that the notification format from `popInitialNotification` is different from the one in `onNotification` because it does not pass through the internal `_onNotification`

teyou:
same issue here.. it was fine if the notification was received during app is in background mode.

Only when app is killed, received a notification message then user click on the notification from notification area, the payload data is not transferred to onNotification callback (as below)

```
{ 
  foreground: false,
  userInteraction: true,
  message: undefined,  // <---  bug???
  data: undefined,  // <---  bug???
  badge: undefined,
  alert: undefined,
  sound: undefined 
}
```

japesh:
This problem can be solved by making some changes in native code.

In RNPushNotification.java file implement **LifecycleEventListener** like this:

public class RNPushNotification extends ReactContextBaseJavaModule implements ActivityEventListener,LifecycleEventListener {
    public RNPushNotification(ReactApplicationContext reactContext) {
        super(reactContext);
  ....
  
        reactContext.addLifecycleEventListener(this);
.... 
  }
 ...
.@Override
    public void onHostResume() {
        Activity activity = getCurrentActivity();
        if (activity != null) {
            Intent intent = activity.getIntent();
            Bundle bundle = intent.getBundleExtra("notification");
            if (bundle != null) {
                bundle.putBoolean("foreground", false);
                bundle.putBoolean("userInteraction", true);
                mJsDelivery.notifyNotification(bundle);
            }
        }
    }

    @Override
    public void onHostPause() {
        // Activity `onPause`
    }

    @Override
    public void onHostDestroy() {
        // Activity `onDestroy`
    }
...
}
That's all and onNotification callback will be triggered everytime

[...]

PR #220, opened 2016-10-04
Run onNotification() in JS, even when the app isn't running at all.

mikelambert:
Make the background listener service start the React JS application (sans Components), and then uses the results of any PushNotification.configure() calls to trigger the JS-based onNotification from the listener service.

This avoids the need to package everything into the Android notification payload to get it to show up, and should hopefully fix a bunch of open issues against this repo.

Note, I haven't explored what's required to get this equivalent functionality working on iOS (and I suspect it may not even be possible at all). But this helps me avoid duplicating functionality on Android and JS by letting me combine it.

I suspect there is additional work that might help combine other native-Android notification processing (RNPushNotificationPublisher and RNPushNotificationBootEventReceiver?) to get them to use JS notification processing. But I don't use them, so they are not in my critical path, and am worried I might break them, since playing with this code is a bit tricky to get right (with threads and the many ways Android notifications can be delivered, etc).

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -9,6 +9,7 @@ import android.content.Intent;
 import android.content.IntentFilter;
 import android.os.Build;
 import android.os.Bundle;
+import android.util.Log;
 
 import com.dieam.reactnativepushnotification.helpers.ApplicationBadgeHelper;
 import com.facebook.react.bridge.ActivityEventListener;
@@ -120,7 +121,7 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         }, intentFilter);
     }
 
-    private void notifyNotification(Bundle bundle) {
+    void notifyNotification(Bundle bundle) {
         String bundleString = convertJSON(bundle);
 
         WritableMap params = Arguments.createMap();
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -7,7 +7,6 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
-import android.os.Build;
 import android.os.Bundle;
 import android.util.Log;
 
@@ -22,21 +21,17 @@ import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
-import com.facebook.react.modules.core.DeviceEventManagerModule;
-
-import org.json.JSONException;
-import org.json.JSONObject;
 
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
-import java.util.Set;
 
 public class RNPushNotification extends ReactContextBaseJavaModule implements ActivityEventListener {
     public static final String LOG_TAG = "RNPushNotification";// all logging should use this tag
 
     private RNPushNotificationHelper mRNPushNotificationHelper;
     private final Random mRandomNumberGenerator = new Random(System.currentTimeMillis());
+    private RNPushNotificationJsDelivery mJsDelivery;
 
     public RNPushNotification(ReactApplicationContext reactContext) {
         super(reactContext);
@@ -44,11 +39,12 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         reactContext.addActivityEventListener(this);
 
         Application applicationContext = (Application) reactContext.getApplicationContext();
+        // The @ReactNative methods use this
         mRNPushNotificationHelper = new RNPushNotificationHelper(applicationContext);
+        // This is used to delivery callbacks to JS
+        mJsDelivery = new RNPushNotificationJsDelivery(reactContext);
 
         registerNotificationsRegistration();
-        registerNotificationsReceiveNotification();
-        registerNotificationsRemoteFetch();
     }
 
     @Override
@@ -63,22 +59,12 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         return constants;
     }
 
-    private void sendEvent(String eventName, Object params) {
-        ReactContext reactContext = getReactApplicationContext();
-
-        if (reactContext.hasActiveCatalystInstance()) {
-            reactContext
-                    .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
-                    .emit(eventName, params);
-        }
-    }
-
     public void onNewIntent(Intent intent) {
         if (intent.hasExtra("notification")) {
             Bundle bundle = intent.getBundleExtra("notification");
             bundle.putBoolean("foreground", false);
             intent.putExtra("notification", bundle);
-            notifyNotification(bundle);
+            mJsDelivery.notifyNotification(bundle);
         }
     }
 
@@ -92,42 +78,9 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
                 WritableMap params = Arguments.createMap();
                 params.putString("deviceToken", token);
 
-                sendEvent("remoteNotificationsRegistered", params);
+                mJsDelivery.sendEvent("remoteNotificationsRegistered", params);
             }
         }, intentFilter);
-    }
-
-    private void registerNotificationsReceiveNotification() {
-        IntentFilter intentFilter = new IntentFilter(getReactApplicationContext().getPackageName() + ".RNPushNotificationReceiveNotification");
-        getReactApplicationContext().registerReceiver(new BroadcastReceiver() {
-            @Override
-            public void onReceive(Context context, Intent intent) {
-                notifyNotification(intent.getBundleExtra("notification"));
-            }
-        }, intentFilter);
-    }
-
-    private void registerNotificationsRemoteFetch() {
-        IntentFilter intentFilter = new IntentFilter(getReactApplicationContext().getPackageName() + ".RNPushNotificationRemoteFetch");
-        getReactApplicationContext().registerReceiver(new BroadcastReceiver() {
-            @Override
-            public void onReceive(Context context, Intent intent) {
-                Bundle bundle = intent.getBundleExtra("notification");
-                String bundleString = convertJSON(bundle);
-                WritableMap params = Arguments.createMap();
-                params.putString("dataJSON", bundleString);
-                sendEvent("remoteFetch", params);
-            }
-        }, intentFilter);
-    }
-
-    void notifyNotification(Bundle bundle) {
-        String bundleString = convertJSON(bundle);
-
-        WritableMap params = Arguments.createMap();
-        params.putString("dataJSON", bundleString);
-
-        sendEvent("remoteNotificationReceived", params);
     }
 
     private void registerNotificationsReceiveNotificationActions(ReadableArray actions) {
@@ -143,7 +96,7 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
                 Bundle bundle = intent.getBundleExtra("notification");
 
                 // Notify the action.
-                notifyNotificationAction(bundle);
+                mJsDelivery.notifyNotificationAction(bundle);
 
                 // Dismiss the notification popup.
                 NotificationManager manager = (NotificationManager) context.getSystemService(context.NOTIFICATION_SERVICE);
@@ -151,32 +104,6 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
                 manager.cancel(notificationID);
             }
         }, intentFilter);
-    }
-
-    private void notifyNotificationAction(Bundle bundle) {
-        String bundleString = convertJSON(bundle);
-
-        WritableMap params = Arguments.createMap();
-        params.putString("dataJSON", bundleString);
-
-        sendEvent("notificationActionReceived", params);
-    }
-
-    private String convertJSON(Bundle bundle) {
-        JSONObject json = new JSONObject();
-        Set<String> keys = bundle.keySet();
-        for (String key : keys) {
-            try {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-                    json.put(key, JSONObject.wrap(bundle.get(key)));
-                } else {
-                    json.put(key, bundle.get(key));
-                }
-            } catch (JSONException e) {
-                return null;
-            }
-        }
-        return json.toString();
     }
 
     @ReactMethod
@@ -218,7 +145,7 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
             Bundle bundle = intent.getBundleExtra("notification");
             if (bundle != null) {
                 bundle.putBoolean("foreground", false);
-                String bundleString = convertJSON(bundle);
+                String bundleString = mJsDelivery.convertJSON(bundle);
                 params.putString("dataJSON", bundleString);
             }
         }
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -137,7 +137,7 @@ public class RNPushNotificationHelper {
             }
 
             if (bundle.getString("message") == null) {
-                Log.e(LOG_TAG, "No message specified for the notification");
+                Log.e(LOG_TAG, "No message specified for the notification: " + bundle);
                 return;
             }
 
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationJsDelivery.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationJsDelivery.java
@@ -1,0 +1,80 @@
+package com.dieam.reactnativepushnotification.modules;
+
+import android.os.Build;
+import android.os.Bundle;
+import android.util.Log;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.Set;
+
+import static com.dieam.reactnativepushnotification.modules.RNPushNotification.LOG_TAG;
+
+/**
+ * Created by lambert on 2016/10/09.
+ */
+
+public class RNPushNotificationJsDelivery {
+    private ReactApplicationContext mReactContext;
+
+    public RNPushNotificationJsDelivery(ReactApplicationContext reactContext) {
+        mReactContext = reactContext;
+    }
+
+    void sendEvent(String eventName, Object params) {
+        if (mReactContext.hasActiveCatalystInstance()) {
+            mReactContext
+                    .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                    .emit(eventName, params);
+        }
+    }
+
+    void notifyRemoteFetch(Bundle bundle) {
+        String bundleString = convertJSON(bundle);
+        WritableMap params = Arguments.createMap();
+        params.putString("dataJSON", bundleString);
+        sendEvent("remoteFetch", params);
+    }
+
+    void notifyNotification(Bundle bundle) {
+        String bundleString = convertJSON(bundle);
+
+        WritableMap params = Arguments.createMap();
+        params.putString("dataJSON", bundleString);
+
+        sendEvent("remoteNotificationReceived", params);
+    }
+
+    void notifyNotificationAction(Bundle bundle) {
+        String bundleString = convertJSON(bundle);
+
+        WritableMap params = Arguments.createMap();
+        params.putString("dataJSON", bundleString);
+
+        sendEvent("notificationActionReceived", params);
+    }
+
+    String convertJSON(Bundle bundle) {
+        JSONObject json = new JSONObject();
+        Set<String> keys = bundle.keySet();
+        for (String key : keys) {
+            try {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+                    json.put(key, JSONObject.wrap(bundle.get(key)));
+                } else {
+                    json.put(key, bundle.get(key));
+                }
+            } catch (JSONException e) {
+                return null;
+            }
+        }
+        return json.toString();
+    }
+
+}
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -2,27 +2,22 @@ package com.dieam.reactnativepushnotification.modules;
 
 import android.app.ActivityManager;
 import android.app.ActivityManager.RunningAppProcessInfo;
-import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.util.Log;
 
 import com.dieam.reactnativepushnotification.helpers.ApplicationBadgeHelper;
-import com.dieam.reactnativepushnotification.modules.RNPushNotification;
-import com.facebook.react.bridge.ReactApplicationContext;
-import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.ReactApplication;
 import com.facebook.react.ReactInstanceManager;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContext;
 import com.google.android.gms.gcm.GcmListenerService;
 
 import org.json.JSONObject;
 
-import java.lang.Runnable;
 import java.util.List;
 import java.util.Random;
-
-import static com.dieam.reactnativepushnotification.modules.RNPushNotification.LOG_TAG;
 
 public class RNPushNotificationListenerService extends GcmListenerService {
 
@@ -59,7 +54,7 @@ public class RNPushNotificationListenerService extends GcmListenerService {
                     // Otherwise wait for construction, then send the notification
                     mReactInstanceManager.addReactInstanceEventListener(new ReactInstanceManager.ReactInstanceEventListener() {
                         public void onReactContextInitialized(ReactContext context) {
-                            sendNotification((ReactApplicationContext)context, bundle);
+                          sendNotification((ReactApplicationContext)context, bundle);
                         }
                     });
                     if (!mReactInstanceManager.hasStartedCreatingInitialContext()) {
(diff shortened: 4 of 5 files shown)


mikelambert:
This should fix:
https://github.com/zo0r/react-native-push-notification/issues/219
https://github.com/zo0r/react-native-push-notification/issues/217
https://github.com/zo0r/react-native-push-notification/issues/170

Though as you mention in https://github.com/zo0r/react-native-push-notification/issues/170#issuecomment-247956844 , I guess it's arguable that you *don't* want this going through onNotification (and instead prefer this native-Java processing system which can't be customized at all), for the sake of iOS compatibility?

I haven't investigated in detail the options detailed in articles like https://medium.com/posts-from-emmerge/ios-push-notification-background-fetch-demystified-7090358bb66e#.krswqcyl6, but I hold out hope that there's some level of background execution that can work on iOS too, and let library users customize how notifications are processed when the app is not running...

npomfret:
Thanks for this, I'll take a look at it later.

logileifs:
@mikelambert I tested your branch but still get no call to onNotification on cold start from notification.

`W/System.err(25916): java.io.IOException: INVALID_SENDER
W/System.err(25916): 	at com.google.android.gms.iid.zzc.zzb(Unknown Source)
W/System.err(25916): 	at com.google.android.gms.iid.zzc.zza(Unknown Source)
W/System.err(25916): 	at com.google.android.gms.iid.InstanceID.zzc(Unknown Source)
W/System.err(25916): 	at com.google.android.gms.iid.InstanceID.getToken(Unknown Source)
W/System.err(25916): 	at com.dieam.reactnativepushnotification.modules.RNPushNotificationRegistrationService.onHandleIntent(RNPushNotificationRegistrationService.java:23)
W/System.err(25916): 	at android.app.IntentService$ServiceHandler.handleMessage(IntentService.java:65)
W/System.err(25916): 	at android.os.Handler.dispatchMessage(Handler.java:102)
W/System.err(25916): 	at android.os.Looper.loop(Looper.java:145)
W/System.err(25916): 	at android.os.HandlerThread.run(HandlerThread.java:61)`

I noticed this was happening when opening the notification and it has something to do with gcm which makes me wonder if this does not work with local notifications ?

mikelambert:
That error looks to be due to the Registration Service, which should be unrelated to what I'm doing (which is processing of notifications), though maybe they interact in strange ways.

In the case of my code, you just have to be sure to call PushNotification.configure(...) *outside* of React.Components, basically called as part of a top-level import, because none of the Components will be initialized in the background thread. Not sure what you're doing exactly, though feel free to link to code/examples of what you're doing and I can take a peek?

npomfret:
I'm just trying to test this, but need some help.  Here's what I've observed so far.

WRT #217 & #219 

My app already receives GCM push notifications when in the background, I don't believe this is a bug.  Maybe there's a bug where it doesn't understand messages that arrive in a different format, but that's a separate issue I think. 

If the app is closed it doesn't receive messages. As far as I can tell from my testing this doesn't address that issue.

WRT #170 

onNotification only gets fired for push notifications (when the app is either in the foreground or background, but not if it isn't running at all).  It does not get fired when the user clicks on an alert in the notification center. And it does not get fired when the app sends a local notification. 

The only change in behaviour I can see is that when my app sends a local notification then onNotification gets called, but not with the expected local notification, but instead it gets called with the last push notification.  So I'm seeing the last push notification over and over again.


In summary, I can't merge this change.  I don't fully understand what its for or how to test it.  Can you please describe exactly what the expected differences in behaviour are please?

mikelambert:
Hey there Nick, sorry for the confusion!

Yes, the existing library already delivers notifications to the JS handler when running in the background. But this patch *adds* support for delivering notifications to the JS handler when the app is closed. I am not sure why it is not working for you in your testing...curious when/how you are calling `PushNotification.configure`?

It does this by loading all the JS from the `ListenerService`, but without constructing any UI components. And then attempting to deliver the notifications using the results of any `PushNotification.configure(...)` calls that happened.

As to the second bit...the "calls onNotification as the result of a localNotification, but using the last push notification" sounds like a bug. I am not sure what's going on here, but can ask a few questions to narrow it down:
- Are you using scheduled notifications or repeating notifications? I have not tested the interactions of those cases.
- Again, where/how are you calling `PushNotification.configure`? Could it be called from a chain initiated from a triggered localNotification? The behavior of repeated calls to `configure({popInitialNotification: true, ...})` has changed, and might cause that. (If we deem this a bug, I can fix this.)
- Otherwise I don't know what is going on, unless somehow local notifications were being delivered through GcmListenerService, which I don't think is supposed to happen from how I understand things.

That said, if you have actual code you are running you can point to (runnable?), that might help me diagnose this better. Push Notifications are kind of a hairy thing to handle, especially with all the JS/Native interactions here, so seeing how things get called in your app might help me diagnose what's wrong. Thanks!

npomfret:
I'm not using scheduled notifications.

My `configure` function gets called when the app is ready to receive push notifications.  So not immediately basically, but after a couple of seconds.

mikelambert:
My question is...what determines when it's "ready to receive notifications".

I feel like I'm not able to communicate myself clearly on the special requirements of configure() to receive JS notifications when the app is not running, since you're not clarifying on how your configure is called. Rephrasing, if someone were to `require('index.android.js')` on your base JS file (without instantiating any of the components, running the constructors, render()s, mounting callbacks, etc), would your `configure()` still be called? It needs to be at the top-level, as an un-indented line, in some imported JS file, for these JS notifications with an closed-app to work.

Or maybe better, can I see any code? Is it online, or a private codebase?

npomfret:
Its a private code base.  My configure method isn't called as you describe, the app starts up and after a while the `configure` function is called.  I think what you're saying is that it has to be called immediately when the app starts?

mikelambert:
Correct. The background listener service is loading and running the JS libraries, but it is not initializing any UI components (since there is no Activity created yet, since the user has not opened the app!)

In my case, I do this:
https://github.com/mikelambert/dancedeets-react/blob/master/js/notifications/setup.js

I call `init()` directly when that code is imported, which sets up an `onNotification` to handle incoming remote notifications.

Later, when the UI is initialized (and I have a redux store available), I call `setup(reduxDispatch)`, and overwrite the `onNotification`, along with setting `popInitialNotification`. This time the onNotification has access to the `dispatch`, and so can control the UI in response to the user opening a notification. (ie, navigate to the notification item the user clicked on). And the `popInitialNotification` ensures the opening-notification gets processed now that the UI has been started properly.

npomfret:
Thanks for the example.  Does your app works on iOS?  When I change my app to call `configure` in this way the `onRegister` function isn't being called.

mikelambert:
I haven't enabled notifications yet for my iOS app, so I haven't tested this. But my changes only affected a small bit of JS, so didn't think it was necessary.

Again, without code, it is hard to debug your situation. What did you change, exactly? Are you passing an onRegister callback to the first, or second, call to configure()? (Or did you only move your call)? Since I haven't changed any of the iOS code, it basically sounds like you are saying:

"There is a bug with the existing react-native-push-notification code on iOS, where it breaks if you call configure too early in the initialization chain"

And so not broken by this pull request...but triggered on iOS because of the constraints this new pull request places on app code calling configure()?

npomfret:
That's correct.  I'll look into it more today

npomfret:
Ah sorry @mikelambert , ignore everything I just said.  I was using the iOS simulator.

npomfret:
Ok, so I've changed my code to call `configure` in index.\<os\>.js, [it's](https://gist.github.com/npomfret/2566cc84ada1a807d201b8146e1fa25e) very similar to yours.  It sets `popInitialNotification` to false, and then set it to true later when the app is ready to process the notification.  I `console.log` each notification as it gets delivered in `onNotification`.

Without adding your change I can get push notifications when the app is backgrounded.  But not when its killed.  I do see the following in the logs though:

```
10-08 14:10:33.840 19416-19416/? D/ActivityThread: BIND_APPLICATION handled : 0 / AppBindData{appInfo=ApplicationInfo{3263f2a com.pomochat}}
10-08 14:10:33.850 19416-19416/? D/ActivityThread: BDC-Calling onReceive: intent=Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x10 pkg=com.pomochat cmp=com.pomochat/com.google.android.gms.gcm.GcmReceiver (has extras) }, receiver=com.google.android.gms.gcm.GcmReceiver@303e11b
10-08 14:10:33.856 19416-19416/? D/ActivityThread: BDC-RECEIVER handled : 0 / ReceiverData{intent=Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x10 pkg=com.pomochat cmp=com.pomochat/com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService (has extras) } packageName=com.pomochat resultCode=-1 resultData=null resultExtras=null}
10-08 14:10:33.881 19416-19416/? D/ActivityThread: SVC-Creating service: CreateServiceData{token=android.os.BinderProxy@36993ccd className=com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService packageName=com.pomochat intent=null}
10-08 14:10:33.881 19416-19416/? D/ActivityThread: SVC-CREATE_SERVICE handled : 0 / CreateServiceData{token=android.os.BinderProxy@36993ccd className=com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService packageName=com.pomochat intent=null}
10-08 14:10:33.881 19416-19416/? D/ActivityThread: SVC-Calling onStartCommand: com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService@3538982, flags=0, startId=1
10-08 14:10:33.883 19416-19416/? D/ActivityThread: SVC-SERVICE_ARGS handled : 0 / ServiceArgsData{token=android.os.BinderProxy@36993ccd startId=1 args=Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x10 pkg=com.pomochat cmp=com.pomochat/com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService (has extras) }}
10-08 14:10:33.940 19416-19449/? E/RNPushNotification: No message specified for the notification
10-08 14:10:33.941 19416-19416/? D/ActivityThread: SVC-Destroying service: com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerService@3538982
10-08 14:10:33.941 19416-19416/? D/ActivityThread: SVC-STOP_SERVICE handled : 0 / android.os.BinderProxy@36993ccd
```

After applying your PR, the change in behaviour when the app is closed is that the notifications are now being delivered to my js code via the `onNotification` function. BUT my app doesn't get started - I don't even see the first `console.log` call from the first `componentWillMount`.  It's as if it's invoking a bit of JS code in my app, but not starting the app.  Is that correct?  It's not the behaviour I was expecting.  I thought it was going to start my app in the background.

So why is `onNotification` even being called  if my app isn't started? I assume that the call to `configure`, which sets `popInitialNotification` to false has happened - so theoretically the notification shouldn't be delivered to my app (yet).  

In addition, when the app is in the background the notification deliveries get repeated.  So the 1st notification received gets delivered once, as expected.  The 2nd gets delivered twice, the 3rd 3 times and so on.  I backed your changes out and retested agains the current npm version to make sure this wasn't my app being screwy but it behaved as expected.

I've looked at the your change, and it seems innocent enough.  I certainly can't see why it's repeat delivering notifications.  I assume you didn't notice this in your testing?

mikelambert:
Great, glad we're getting closer on this.

Yes, your understanding is correct. If the app is not open, a notification will *NOT* start the app UI. That's impossible to do. What it does do is load the JS, in the background Listener Service. That's why I keep reiterating that any code in UI (including `componentWillMount`) will not be called on these notifications received while the app is closed. This is the standard Android notification setup...just applied within React Native.

`popInitialNotification` causes the current *activity*'s (ie, the UI's) current notification to be returned. This is the notification that started the activity, because the user opened a notification (or the notification that was delivered to a backgrounded app that brought it to the foreground). But this should not be the remote notification being delivered to an app (foreground or background, started or unstarted), because that's not associated with the activity/UI.

As far as duplicate notifications, that definitely sounds like a bug. :/ I may have seen something a little similar when operating in hot-reload mode, where (I was guessing) maybe reloads were causing multiple instances of a push handler to be installed. I didn't notice it on a fresh app start/install, or on release mode apps, though.

Just to clarify, when you said the second notification gets delivered twice, what do you mean by "delivered"? The `onNotification` handler gets called twice? What is your notification handler doing...is it possibly calling PushNotification.configure at all? Trying to see if there are any clues that might help me track this down.

npomfret:
Exactly, I'm getting repeated calls to `onNotification`.  I've not debugged it to see where they're coming from yet.  I don't believe it's reloaded to hot reload (I'm sure I tested that theory at the time).  The handler isn't doing anything weird, and the behaviour of it is the same whether or not the app is backgrounded or not - and the repeat calls to `onNotification` only seemed to happen when backgrounded.

npomfret:
I think the bug lies in the RNPushNotification constructor, or how it's being called.  Inside it calls 3 register methods:

```
        registerNotificationsRegistration();
        registerNotificationsReceiveNotification();
        registerNotificationsRemoteFetch();
```

And in turn they add state to the react context.  Now this ctor is being called after a push notification is received in `RNPushNotificationListenerService.onMessageReceived` (which then calls `sendNotification`) if the app is in the background.

```
        if (!isRunning) {
            // Run the notification on the JS thread
            RNPushNotification pushNotification = new RNPushNotification(context);
            pushNotification.notifyNotification(bundle);
        }
```

So with each push request that happens when backgrounded, we are adding more and more receivers for RNPushNotificationReceiveNotification.  So the react context has increasing number of receivers which explains why I'm seeing more and more repeated calls to `onNotification`. 

Before your change this read

        if (!isRunning) {
            new RNPushNotificationHelper(getApplication()).sendNotification(bundle);
        }

So maybe this is where it's gone wrong?  To be honest I'm struggling to follow the code paths and finding it quite confusing.

mikelambert:
Thanks, great debugging, I think that's exactly what's going wrong. (My assumption of hot-reloading was probably unrelated, and I was probably conflating "reload the code to fix a notification bug" with the second notification I sent after reloading.)

Anyways, I've just pushed a fix in a separate change, which you can take a look at directly. Basically it factors out the deliver-event-to-JS functionality from RNPushNotification.java, so that it can be used by the ListenerService directly, without the accidental installation of handlers (and without the need for the extra Intent communication that existed before).

Hopefully this helps clarify the code a bit, too.

npomfret:
Ok, great.  I've retested and it looks fine - thank you!  I'll merge it now.

Forgive me for being stupid, but I'm still a little confused about the purpose of this change.  If your app isn't running, and doesn't get started when the notification arrives - what's this change for?  What can you do if your app isn't running?

slestang:
This change, if I understood it correctly, should allow us to add a 'snooze notification' feature to our app.
We will try it today and give feedback.

mikelambert:
In my case, it allows me to:
- handle incoming notifications with the same code flow, regardless of
whether the app is not running or backgrounded or foregrounded. Otherwise
it makes it really difficult to ensure consistent notifications between
backgrounded and not running instances of the app
- send a data payload to the client, for the client to handle and process
itself. This lets my client localize the notification for the user based on
the phone's language, for example

One could also theoretically use this in a chat application, to track
whether the chat message had been delivered to the remote chat client (a la
Whatsapp and Facebook status indicators)

2016年10月10日(月) 5:41 Silvère Lestang <notifications@github.com>:

This change, if I understood it correctly, should allow us to add a 'snooze
notification' feature to our app.
We will try it today and give feedback.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/zo0r/react-native-push-notification/pull/220#issuecomment-252573625>,
or mute the thread

npomfret:
Hi @ddobby94 @timhonders @jitenderchand1 @JonesN7 @heminm @vikeri  @barroslee  

This change is in master now.  Are you able to try it out before we do a release?  (You'll need to change your _package.json_):

From:
`    "react-native-push-notification": "2.1.1",`

To:
`    "react-native-push-notification": "git@github.com:zo0r/react-native-push-notification.git",`

chaix-adrien:
` "react-native-push-notification": "git@github.com:zo0r/react-native-push-notification.git",`
doesn't work for me, but
`    "react-native-push-notification": "zo0r/react-native-push-notification.git",` 
do the job.

chaix-adrien:
Modifications works fine for me.
The onNotification function is called even if the app is killed.

Thanks !

vikeri:
For me onNotification is only called when I click on the notification. This both works when the app is active and when it is off. When the app is off it starts the app and runs the callback. Was the intention that the callback should run without having to click the notification?

mikelambert:
@vikeri: In theory, the original code should have *also* called `onNotification` when a remote notification was received from the server, and your app was in the foreground or background. This pull request extends that, to also run `onNotification` when the app is not running.

If you check `userInteraction` on the notification passed to `onNotification`, you can tell if the notification was received from the remote server, or actually clicked-and-opened.

vikeri:
@mikelambert Alright, so the onNotification is only triggered when I have a remote notification that arrives? I can not trigger it when a local notification arrives (without clicking it)?

mikelambert:
I believe that's correct.

A local notification is necessarily being sent from within your own app, right? So unless I'm missing something, you shouldn't need to trigger onNotification, since you are executing the code that is sending the local notification, right?

codepujan:
@mikelambert  I modified my package.json file but still I am not getting my onNotification() to fire  after the app is closed.  it works fine while the app is open . Don't know what did went wrong .

npomfret:
Are you testing with a remote push notification?  Or a scheduled local notification?

codepujan:
@npomfret I tried with a remote push notification .

mikelambert:
@codepujan I recommend reading the full thread above, where I went through this problem with @npomfret on why remote push notifications were not being received, and the nature of how code is initialized. I suspect it will explain your situation as well.

Namely, are you calling PushNotification.configure() directly from an imported module, or are you calling it from a React.Component method? You need to be doing the former, as demonstrated in my example link above.

npomfret:
Yes I forgot about @mikelambert 's point.  I had to refactor my app a little to get it to run some js code when it had been killed and a push notification arrived.  Basically i needed to invoke the call to `configure` outside of any of the react-native life-cycle.  Just do it in a js file that gets imported by your index.os.js file. Something like [this](https://gist.github.com/npomfret/2566cc84ada1a807d201b8146e1fa25e).

codepujan:
@npomfret  +1 . It's working . You guys have done a great work. Kudos

jitenderchand1:
I moved my code to index.android.js file. onNotification callback is working fine but my onRegister function stop working.


```
PushNotification.configure({
                onRegister: async (token)=> {
                console.log(token)
 },
requestPermissions:true,
senderID:"31**********",
popInitialNotification:true
}
```

bo4a10:
Hi guys!! @npomfret @mikelambert, I changed my packege.json to 
"react-native-push-notification": "zo0r/react-native-push-notification.git".
Now onNotification handler invokes in any cases when remote PN arrived, BUT...
every remote PN didn't trigger sound and for end-user it looks like app didn't receive PN at all.
Is the idea of your aproach is to receive every remote PN as 'silent' PN and then generate Local PN (when it is reuired to show PN for user)?!

Thanks in andvance!

bo4a10:
Seems like onNotification handler doesnt fires when app closed and user click on the local notification.

mikelambert:
@jitenderchand1 Sorry not sure what's up there, did you ever figure it out? I have `PushNotification.configure`called directly from an import from `index.android.js`, and it works ok for me AFAIK. (And yes, I'm using an async function as well)

@bo4a10 
It should be the same whether the app is open or closed. My understanding is this: When a notification is received when the app is open, the app's `onNotify` is called. It is the job of `onNotify` to construct a local notification if desired. When a notification is clicked to open, onNotify is called again (this time with `userInteraction === true`).

So if the app is closed, it is still the responsibility of your app to present the local notification. This should be the same as when the app is open, right?

When the app is closed, and the user clicks on the notification, it starts your app with a pending notification. This will not call `onNotify` automatically, unless you set `popInitialNotification` in your call to `configure` (to have it called automatically), or call `PushNotification.popInitialNotification()` directly (to trigger a call).

bo4a10:
@mikelambert, thanks for response! Yes, you are right, and I expecting the flow like you are described. I called `PN.configure` with `onNotification` handler and `popInitialNotification` setted to false when import script in index.os file, and then in index.os React Component constructor I call `PN.configure` with `popInitialNotification` setted to true... Probably everything like you and @npomfret described here above.

Everything is clear for me except situation when app closed completly, and user will click on local notification that I created after I received remote notification... app just opens, and nothing triggers or 'says' me smth like 'hey, bro, you opened app by clicking on this local push notification'. :)

mikelambert:
I'm not sure. That situation is unrelated to the changes I've made, and you'd have that problem with any notification that triggers your app to open. And yes, `popInitialNotification` *should* handle it. And it works for me, doing exactly what you described above.

My only suggestion would be to dive in and start logging/debugging. In particular, what is returned to `getInitialNotification`'s callback in `component/index.android.js`, and where do things break down on the passing-up to `onNotification`? Of if nothing is returned, then dig into the java: What happens inside `RNPushNotification.java`'s `getInitialNotification`, and what does it return? Is there any problems handling the Intent...what `intent-filter`s are listed for your `AndroidManifest.xml`'s `MainActivity`, etc.

bo4a10:
ok, thanks for help @mikelambert, I'll investigate it deeper.
btw, one more question: should end-user see remote notifications now(with your changes) or they are don't appear in status bar anymore and end-user can't see them(we can catch them just in `onNotification`)?

I asked because when PN module in my package.json was in 2.0.1 version - remote notifications appears in status bar with sound and so on (they were visible for end-user), but when I checked out to master version of PN module(for closed app purposes) I can catch remote notification just in `onNotification` handler.

Hope that your answer will help me to understand on what point something went wrong, just on clicking on local notifications event or somewhere in global configs of the module.

Thanks in advance!

mikelambert:
My personal usage of this library is to process remote notifications, and then produce actual local notifications with all the necessary field.

Previously, I believe if you passed all the necessary fields in the remote notification that were necessary for display, then the remote notification would have been presented to the user. (This was never documented here though, I only found it when digging in to the code...) This would have happened for app-loaded or app-closed scenarios, from what I understood. Is this what you observed? (Or did you only see this behavior work for app-loaded scenarios, in which case I'm confused?)

But yes, it's possible I broke the old undocumented behavior, since I don't use it.. :/

bo4a10:
Hi @mikelambert. Seems like I found solution for my issue...
`popInitialNotitifaction` doesn't override when I initialize `PN.configure` at second time(in Component constructor) because `PN.isLoaded` property setted to true after first initializing(for closed app PNs).
`/react-native-push-notification/index.js` line 78.

I dont think it must be so, but my solution is to set `PN.isLoaded = false` before initialization of PN module in component constructor...

mikelambert:
Ahh thanks for debugging, though I'm a bit confused as to what's actually not working here. (I'd like to ensure we fix it, if there's a bug...)

In my pull request, I had actually split out the code into both `this.isLoaded` (which installed handlers) and `this.hasPoppedInitialNotification` (which ensured we only popped a single time).

With `isLoaded==true`, as long as you haven't already passed `popInitialNotification` before in your first call, it should still look at `popInitialNotification` in your second call to determine whether to pop and call your `onNotification`.

So I'm confused as to what line of code inside the `configure()` call is being run when you manually set `PushNotification.isLoaded=false`, that allows your code to work?

npomfret:
Is it related to this recently merged PR https://github.com/zo0r/react-native-push-notification/pull/242 ?

bo4a10:
Guys, I'm really sorry! :)
@mikelambert, your last message saved situation. I found that I havn't got this part of code `this.hasPoppedInitialNotification` in index.js of the module. And I found that another dev replaced index.js file to his own modified file before building app...
Now I fixed this and `onNotification` fires everry time remote notification arrived or I pressed on local notification when app completely closed. So, seems like now everything works as you and probably I expected to work :)

Perfect job @mikelambert . Sorry and thank you a lot for help again. :)

@npomfret, this my 'issue' was not related to #242. I don't use `unregister()` method - I have no need for it. thanks.

npomfret:
@zo0r i think we're ready for a release.

zo0r:
@npomfret published! Drop me an email to z@zo0r.me if you want npm access too.

stu60610:
Hi All, I've follow your steps to get the push message when app is completely closed, and when I pressed on notification, the app start and call `onNotification` successfully, but the notification object's message & data are all undefined.(like #161) Did I miss something to setup?

chirag04:
@mikelambert our build started failed when updating from v2.1 to v2.2 of this lib. Turns out this commit breaks it for us. (Note we are building RN from source).

Below the the log. Would you have any idea?

```
:react-native-push-notification:compileReleaseRenderscript UP-TO-DATE
:react-native-push-notification:generateReleaseBuildConfig UP-TO-DATE
:react-native-push-notification:generateReleaseResValues UP-TO-DATE
:react-native-push-notification:generateReleaseResources UP-TO-DATE
:react-native-push-notification:mergeReleaseResources UP-TO-DATE
:react-native-push-notification:processReleaseManifest UP-TO-DATE
:react-native-push-notification:processReleaseResources UP-TO-DATE
:react-native-push-notification:generateReleaseSources UP-TO-DATE
:react-native-push-notification:incrementalReleaseJavaCompilationSafeguard UP-TO-DATE
:react-native-push-notification:compileReleaseJavaWithJavac
:react-native-push-notification:compileReleaseJavaWithJavac - is not incremental (e.g. outputs have changed, no previous execution, etc.).
/Users/chirag/Desktop/qsi/comlink/node_modules/react-native-push-notification/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java:12: error: cannot find symbol
import com.facebook.react.ReactApplication;
                         ^
  symbol:   class ReactApplication
  location: package com.facebook.react
/Users/chirag/Desktop/qsi/comlink/node_modules/react-native-push-notification/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java:59: error: cannot find symbol
                ReactInstanceManager mReactInstanceManager = ((ReactApplication) getApplication()).getReactNativeHost().getReactInstanceManager();
                                                               ^
  symbol: class ReactApplication
Note: /Users/chirag/Desktop/qsi/comlink/node_modules/react-native-push-notification/android/src/main/java/com/dieam/reactnativepushnotification/helpers/ApplicationBadgeHelper.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
2 errors
:react-native-push-notification:compileReleaseJavaWithJavac FAILED

FAILURE: Build failed with an exception.
```

kanerogers:
@chirag04 I am having the same issue. Did you get any fixes?

robcalcroft:
I can receive remote notifications fine even from cold start but onNotification isnt called in the background.

jmeiss:
Me neither. Do you still have the issue?

nonameolsson:
@robcalcroft did you solve this?

vvavepacket:
@mikelambert 

I have a question regarding your _setup(reduxDispatch)_

My situation is, so react native app is closed, and when a notification comes in, the notification kicks in, but it can't execute the payload which is navigating thru different UI screens by dispatching some actions.. I noticed that when the app is closed and a notification is received, the UI is not initialized, so those dispatch actions that were sent from onNotification were ignored.

So, when the user tries to click the notification, my current behavior is it launches the app and points to the default screen. **How can we re dispatch, or reexecute the actions that was received?** I tried looking in your repo https://github.com/mikelambert/dancedeets-react but I could'nt find any part of the code calling _setup(reduxDispatch)_

anton96:
@vvavepacket Did you solve your issue?

Issue #56, opened 2016-04-29
Crash: JSON value of type NSDate cannot be converted to a date


ammichael:
Hello!

I'm scheduling multiple notifications, but sometimes (that's the worst part; it's just sometimes) it crashes and throws this error:

`JSON value '2016-05-25 04:00:00 +0000' of type NSDate cannot be converted to a date`

Here is the code used to schedule:

```
static scheduleNotification(quoteDate, quoteHours, quoteOfDay){
	let notifDate = new Date(quoteDate);
	notifDate.setHours(quoteHours.hour, quoteHours.mins, 0);
	let message = '"'+quoteOfDay.frase + '" - ' + quoteOfDay.autor;
	PushNotification.localNotificationSchedule({
		id: 0,
		message: message,
		date: notifDate
	})
}
```

I've tried to use "date: new Date(notifDate)", but no lucky.
Any ideas?

Thank you

Edit:

It seems to be related with this issue: https://github.com/facebook/react-native/issues/4049

ammichael:
Hey guys, anyone having this problem, it might be solved passing a date in ISO 8601 format. Just use ".toISOString()". That works for me, all the crashes are gone! (:

mtolly:
You can also pass the unix timestamp number from `.getTime()`.

ammichael:
Yeah, probably it would work too. I exhaustively tested with ISOString and had no problems, not sure about unix timestamp, but it makes sense.

igrayson:
None of the approaches here seem to work for a multi-platform usage.

[`localNotificationSchedule'](https://github.com/zo0r/react-native-push-notification/blob/master/index.js#L134) expects `date.getTime()` to exist, as it would when used as the README describes.

superandrew213:
For Android use: 
```
date: new Date(Date.now() + (10 * 1000))
```
for iOS use:
```
date: new Date(Date.now() + (10 * 1000)).getTime()
```

Is there a reason why we are doing this? 
```
Notifications.localNotificationSchedule = function(details: Object) {
	if ( Platform.OS === 'ios' ) {
		this.handler.scheduleLocalNotification({
			fireDate: details.date,
			alertBody: details.message
		});
	} else {
		details.fireDate = details.date.getTime();
		delete details.date;
		this.handler.scheduleLocalNotification(details);
	}
};
```

PR #228, opened 2016-10-09
fixed scheduleLocalNotification to work on iOS

marciobueno1:
This has been addressed here: #56

I added the .toISOString(), so there will be no error when scheduling notifications on iOS.


npomfret:
what error were you seeing?  I need to be able to reproduce the bug before applying your PR so I can test the change in behaviour

npomfret:
Thank you

marciobueno1:
@npomfret Thanks for merging this, and sorry for not answering you before.

andwaal:
A note to the maintainers of this project, this is kinda a breaking change. 
In our case we had coded around this issue by having separate modules for iOS/Android as recommended in #56, but after this was merged that broke.

PR #242, opened 2016-10-19
set isLoaded to false when unregister

ianlin:
When I call `unregister()` and then call `configure()`, since `this.isLoaded` is true already the event listeners will not be set. So the `_onRegister()` and `_onNotification()` callback will not be called either. This PR fixes this by setting `this.isLoaded` to false when calling `unregister()`.


npomfret:
Thanks for this.

Can you explain how I would go about testing?  What steps would I need to take to observe the broken behaviour?

ianlin:
My use case was:

- The user sign in
- Register push token by calling `configure()`
- Receive push token via `onRegister()` callback and send it to our server
- User can receive push notification happily
- User sign out the app
- The sign out process calls `unregister()` - note that at this point the app is still not killed
- User sign in again
- Calls `configure()` but never receive `onRegister()` callback

Let me know if there's anything not clear, cheers!

PR #247, opened 2016-10-22
Fix doc

varungupta85:
This adds documentation for notification actions and repeating notifications and also some comments requested by @npomfret


npomfret:
thanks

PR #266, opened 2016-11-05
Use not default but passed soundName when playSound is not passed

tsugitta:
If `playSound` option is not passed, soundName option is overrided by `'default'`.
therefore, for now, if we want to change `soundName`, we have to pass `playSound: true` explicitly.


npomfret:
Hi - thanks for this.  Are you able to fix the conflicts and I'll look into testing it

tsugitta:
@npomfret 
fixed the conflict. please check.

npomfret:
@zo0r this and a couple of other things have been merged to master.  After a couple more people have tried it out I think we're ready for a release

zo0r:
@npomfret `2.2.1` published.

PR #269, opened 2016-11-07
[Android] Remote notifications should be added to the notifications centre when app is not running

kfiroo:
Call RNPushNotificationHelper.sendNotification when a notification arrives and app is not running to present a notification to the user

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -1,5 +1,6 @@
 package com.dieam.reactnativepushnotification.modules;
 
+import android.app.Application;
 import android.app.ActivityManager;
 import android.app.ActivityManager.RunningAppProcessInfo;
 import android.os.Bundle;
@@ -19,6 +20,8 @@ import org.json.JSONObject;
 import java.util.List;
 import java.util.Random;
 
+import static com.dieam.reactnativepushnotification.modules.RNPushNotification.LOG_TAG;
+
 public class RNPushNotificationListenerService extends GcmListenerService {
 
     @Override
@@ -37,6 +40,8 @@ public class RNPushNotificationListenerService extends GcmListenerService {
                 ApplicationBadgeHelper.INSTANCE.setApplicationIconBadgeNumber(this, badge);
             }
         }
+
+        Log.e(LOG_TAG, "RNPushNotificationListenerService.onMessageReceived: " + bundle);
 
         // We need to run this on the main thread, as the React code assumes that is true.
         // Namely, DevServerHelper constructs a Handler() without a Looper, which triggers:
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
```diff
@@ -93,6 +98,14 @@ private void sendNotification(ReactApplicationContext context, Bundle bundle) {
         if (bundle.getString("contentAvailable", "false").equalsIgnoreCase("true")) {
             jsDelivery.notifyRemoteFetch(bundle);
         }
+
+        Log.e(LOG_TAG, "RNPushNotificationListenerService.sendNotification: " + bundle);
+
+        if (!isRunning) {
```
mikelambert: This only triggers if the app is not running. So an app in the foreground, it's the responsibility of the client JS code to present a notification (or in-app display, right?) What about if the app is running in the background where there is no in-app display to show it...this change will still not deliver it to the notification center. Is that desired? (Sorry, I'm a bit unclear on what the behavior used to be.)

kfiroo: @mikelambert I was simply restoring the old behaviour that was removed, as it cause a huge regression in my app. Our android app stopped showing notifications.

If you ask me we should investigate exactly how `PushNotificationIOS` behaves and implement the same.

I think this is it, but I'll take another look just to make sure and let you know.

mikelambert: Ahh okay I dug a bit more now that I'm back at a computer, your code is the same as it was before my change.

isRunning and isApplicationRunning both check if the app is *foregrounded*, not merely that it is running. For the sake of better clarity of the old code, can you s/isApplicationRunning/isApplicationInForeground/ and s/isRunning/isForeground/?

Otherwise this change is a valid fix and I am fine with merging (though lack the power) to fix the issues people are seeing.

npomfret: I can merge it if you guys are happy. The rename sounds like a good idea though

kfiroo:
@npomfret @mikelambert Done

npomfret:
Thanks - i'll try it out in the morning

npomfret:
Ok, that's done.  I changed some logging, renamed a couple of methods but made no functional changes.

thanks guys

georgi-kovachev:
when do you think this feature will be added?
I think this one is not merged in current version. 2.2.0? right?

npomfret:
It's in master.  Would you mind trying it out?

teyou:
thanks, this fixes the breaking changes I'm having after upgraded from version 2.0.2.
Now i'm getting the notification no matter the app is on background or killed (the same behaviour as 2.0.2). Hence, I don't need to manually trigger "localNotification" anymore.

JohnAntonusMaximus:
On version 3.0.1 and still getting this problem when app is started but running in background on Android (on the home screen). As soon as a notification comes in, the I get 'Unfortunately <App Name> has stopped working'

PR #288, opened 2016-11-16
Add repeatType for IOS

Jude229:
Check this comment explaining well, what the aim of this pull is. [Thanks buddy](https://github.com/zo0r/react-native-push-notification/pull/288#issuecomment-274985839) @dbhowell

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,7 +158,8 @@ Notifications.localNotificationSchedule = function(details: Object) {
 		this.handler.scheduleLocalNotification({
 			fireDate: details.date.toISOString(),
 			alertBody: details.message,
-			userInfo: details.userInfo
+			userInfo: details.userInfo,
+			repeatInterval: details.repeatType
 		});
 	} else {
 		details.fireDate = details.date.getTime();


npomfret:
Thanks for this.

Can you describe what the change in behaviour is and what step I need to take in order to test it?

Jude229:
Heyy @npomfret,

Well to test it you'll need react-native 0.37 as the repeatInterval was introduced in that version.

There is no change in the behaviour, you use the library as u do for android. The main difference is that you can not use `repeatType : 'time'`for IOS.

Hope you got the answers you were looking for. :simple_smile:

ippy04:
👍

npomfret:

dbhowell:
Hi @npomfret, while I'm not the PR author, I think what they are trying to say is:

* The `PushNotificationIOS` now supports repeating intervals via the `repeatInterval` property. (see https://facebook.github.io/react-native/docs/pushnotificationios.html#schedulelocalnotification)
* `react-native-push-notification` uses `repeatType` for Android, so is not compatible with `PushNotificationIOS`.

This PR aliases `repeatType` to `repeatInterval` for IOS scheduled notifications.

It may be better to match the `PushNotificationIOS` properties instead of `react-native-push-notification` for better support from Developers who are familiar with `PushNotificationIOS`?

Either way I suggest this is a good PR to merge right now.

Jude229:
@dbhowell thanks that's exactly what I meant. @npomfret where should I put the description then ?

npomfret:
A description in this discussion is fine.  Once 2 people have checked a PR works and everyone is in agreement that the code is in good shape and the purpose of the PR is good, then we're ready to merge.  

I think having our own names for the concepts here isn't a bad idea (rather than using the iOS names).  It gives us a little abstraction that could be useful.  I think this is approach that's been taken so far elsewhere in the project. (although I can't think of an example right now - its late, sorry).

eujern:
This looks good to me, though I think it's better to handle and delete invalid values rather than relying RTCPushNotificationManager.m to handle it.  

Also agree with having a different name for repeatType, though it would be a break backward compatibility.

npomfret:
Breaking compatibility is ok in my opinion, that's the the major version is for in npm.

rikur:
⏳

onlinegeek101:
I am currently just pointing at Jude's branch. I also am going to add a PR that allows you to just send custom values from this API so for cases like this you always have a work around to just manually set the desired RN value.

abarisic86:
This isn't merged? Documentation says it doesn't support the iOS yet.
There is a PR that enables requested feature: https://github.com/zo0r/react-native-push-notification/pull/381
Please go there and add 👍 so that it draws a bit of attention.

PR #329, opened 2017-01-09
Add usesChronometer to android notification.

ftufek:
Displays timer in the Android notification through the usesChronometer option. Tested with local app.


varungupta85:
hi @ftufek, @npomfret, one thing to note while reviewing any PRs is that whenever a new notification attribute is added for Android, it should also be added to class `RNPushNotificationAttributes` or else that attribute would be lost if the notifications are restored after a reboot.

varungupta85:
@ftufek Also, I think it would help if you could post some snapshot of how the chronometer looks in the Android notification. Thanks!

ftufek:
I'll do that change and create another pull request from a different branch. Thanks!

PR #336, opened 2017-01-16
Feature: cancel a scheduled notification by its ID

kenma9123:
This feature is created because `cancelLocalNotifications` does not work at all for some reason.

So to make it simple, just pass the id of the notification to remove from the notification centre.
All you have to do is
```
const id = "NOTIFICATION_ID";
PushNotification.cancelLocalNotificationById(id);
```

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -183,6 +183,14 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
 
     @ReactMethod
     /**
+     * Cancel a scheduled notifications by Id and removes notification from the notification centre.
+     */
+    public void cancelLocalNotificationById(String notificationIDString) {
+        mRNPushNotificationHelper.cancelScheduledNotificationById(notificationIDString);
+    }
+
+    @ReactMethod
+    /**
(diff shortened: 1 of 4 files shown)


npomfret:
Thanks for the work but I'd rather figure out why the existing implementation doesn't work.  The reason is that this API should, where possible, mirror the iOS api.  And this is one feature where it should be possible.

kenma9123:
Great, but at least somehow improve the documentation. I tried what's on the documentation but it doesn't really work at all. So until then I'll stay with my PR to remove the notifications. Thanks.

npomfret:

superandrew213:
@npomfret @kenma9123 the issue is [here](https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L423)

You only need to match ids
```
        if (notificationAttributes.getId().equals(userInfo.getString("id"))) {
            cancelScheduledNotification(id);
        }
```

See my [PR](https://github.com/zo0r/react-native-push-notification/pull/371)

KatSick:
Is there any updates ? Canceling local scheduled notifications not working on Android.. on iOS works like a charm

nonameolsson:
@KatSick Did solve this? Does it work for you?

mahsabeyghole1:
i use push notification schadule but for count of start my app send notification on specific time.
how to solve it?

PR #339, opened 2017-01-19
Enable Push on Android < 5.0 (less then API 21)

Hitabis:


Hitabis:
On older android version the foreground property was always returning true even if the app was running in background.

npomfret:

yianyue:
For us, this PR fixed a bug for Android sdk < 21 where notifications do not show up when app is in the background but not closed. The notification object had `foreground = true`.

We did need to add `<uses-permission android:name="android.permission.GET_TASKS" android:maxSdkVersion="20" />` to the android manifest.

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
```diff
@@ -115,7 +123,12 @@ private void handleRemotePushNotification(ReactApplicationContext context, Bundl
     }
 
     private boolean isApplicationInForeground() {
+        if(VERSION.SDK_INT < 21){
+            return appInForegroundForOldApi();
+        }
         ActivityManager activityManager = (ActivityManager) this.getSystemService(ACTIVITY_SERVICE);
+
+        List<ActivityManager.RunningServiceInfo> serviceInfos = activityManager.getRunningServices(Integer.MAX_VALUE);
```
varungupta85: You are not using `serviceInfos` anywhere. Is this required?

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
```diff
@@ -7,6 +7,11 @@
 import android.os.Handler;
 import android.os.Looper;
 import android.util.Log;
+import android.os.Build.VERSION;
+import android.app.usage.UsageStatsManager;
+import android.app.usage.UsageStats;
+import android.content.Context;
+
```
varungupta85: I don't think any of the extra imports are also required except `import android.os.Build.VERSION;`.

PR #358, opened 2017-02-08
Support for nested notification bundle

sebo:
I couldn't edit the existing Pull Request https://github.com/zo0r/react-native-push-notification/pull/86 so I created this. Also the changes are less significant and real easy to review.

Currently you are not supporting nested notification bundles like already described in here https://github.com/zo0r/react-native-push-notification/pull/86

In the "new" gcm/firebase version of Android its possible to send data and notification json. To access this data in react native this library needs to also parse nested json separately since JSONObject.wrap doesn't support the bundle datatype. I added it to the existing convertJSON by changing convertJSON to convertJSONNested and the existing convertJSON using the newly created convertJSONNested. Please have a look.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -141,8 +141,9 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         Activity activity = getCurrentActivity();
         if (activity != null) {
             Intent intent = activity.getIntent();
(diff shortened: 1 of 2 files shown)


npomfret:
Thanks for your PR - all contributions are greatfullly received!

Could you find someone to test this?  Perhaps another person who has an outstanding PR would would help (and in turn you test theirs).  Please detail

 * how to test it
 * what to test on (physical device, emulator, simulator)
 * what operating systems are affected plus any OS version requirements
 * what is being fixed (or added)
 * what is the current behaviour, and crucially - **what is the expected change in behaviour**

Dallas62:
This PR is already implemented in codebase.

Issue #294, opened 2016-11-23
Actions not responding after backing out of app on Android


alexduckmanton:
Hey there, 

I've got some local notifications with an action running for an audio app. The actions work great when the app is running (including when in the background), but they don't seem to respond when the app's been closed. Specifically, the action should stop audio playback. 

Tapping the notification itself will re-launch the app, which is fine, but tapping an action just clears the notification. The action itself fires some code in RN, which is then passed over to some native code to control playback.

So if there isn't a way to fire that RN code from an action when the app is closed, is there a way for an action to be tied directly to some native code instead?

Thanks!

jaisonveneri:
Same here +1

sticknor:
+1

varungupta85:
Hi, I am not sure how would the audio play if the app is completely closed. But you can definitely control some of the actions in the native code directly. It is not there in this package for obvious reasons but you can check out how are the actions handled right now and you will get a good idea of how you can achieve it if you are familiar with Java. Basically, you can add that action to the notification in the js code and then in the native code instead of firing the notification action event for such actions, handle them directly right there. HTH

varungupta85:
Although the ability to handle notification actions when the app is not running is still an issue which I am also having at the moment and not able to spend time to find a way to fix it. I guess we want the app to start on receiving a notification of it is not currently running and make sure that the notification actions are initialized without needing any component to be mounted. A commit from @mlambert did add such functionality but I don't have that code in my fork since it doesn't play nice with one of the packages that I am using.

EdenGottlieb:
+1. My app suffers from this as I rely on an ongoing notification with actions, we need some sort of solution

npomfret:
I've had this working, there is some info in the trouble shooting guide.  

You need to code your app in such a way that the RN-push-notifications `PushNotification.configure..` code gets called __outside__ of the react-native life-cycle, basically as a side effect of an import in your top level _index.js_ file.  The reason is that it seems that when android starts your app in the background from cold as a result of receiving a remote push notification, it doesn't appear to start the react-native life-cycle.  So code in your app will run, but non of the `componentDid/WillMount` methods will get called.

It might be a significant code change to get this working.  I've had to essentially create a _headless_ js app with no react-native in it at all, then I just use react-native as a display mechanism.

jaisonveneri:
Hello @npomfret,

Is this very complex? Do you have any documentation?

Thank you.

npomfret:
[...]

PR #359, opened 2017-02-09
check if it’s a remote notification and pass all the data transferred…

sebo:
"notification" is actually not set with google remote messages if you open the app after a click on the notification on android, so we need to check for google.message_id instead and then transfer the properties send with the data object, so that they can be read inside of the app.


joaovpmamede:
@sebo Is this for when the app is closed? Possibly [related](https://github.com/zo0r/react-native-push-notification/issues/294#issuecomment-286771460).

npomfret:
Could you find someone to test this?  Perhaps another person who has an outstanding PR would would help (and in turn you test theirs).

Issue #368, opened 2017-02-15
cancelLocalNotification not working


luifer-villalba:
I have an issue trying to figure out how to make it work with `cancelLocalNotification`.

This is the local notification definition. Currently, working for Android.
```
PushNotification.localNotification({
      /* Android Only Properties */
      id: '1',
      ticker: 'Notification Center', // (optional)
      autoCancel: true, // (optional) default: true
      largeIcon: 'ic_launcher', // (optional) default: "ic_launcher"
      smallIcon: 'ic_notification', // (optional) default: "ic_notification" with fallback for "ic_launcher"
      subText: 'Alert', // (optional) default: none
      color: 'red', // (optional) default: system default
      vibrate: true, // (optional) default: true
      vibration: 300, // vibration length in milliseconds, ignored if vibrate=false, default: 1000
      tag: notification.type, // (optional) add tag to message
      group: 'group', // (optional) add group to message
      ongoing: false, // (optional) set whether this is an "ongoing" notification
      lat: geolocation.lat,
      lon: geolocation.lon,
      title: 'ALERT!', // (optional, for iOS this is only used in apple watch, the title will be the app name on other iOS devices)
      message: notification.message, // (required)
      playSound: false, // (optional) default: true
      soundName: 'default', // (optional) Sound to play when the notification is shown. Value of 'default' plays the default sound. It can be set to a custom sound such as 'android.resource://com.xyz/raw/my_sound'. It will look for the 'my_sound' audio file in 'res/raw' directory and play it. default: 'default' (default sound is played)
      actions: '["Yes", "No"]'  // (Android only) See the doc for notification actions to know more
    })
```

`cancelAllLocalNotifications` works but I don't want to lose all my local notifications.

This is how I'm trying to cancel by attribute my local notifications `PushNotification.cancelLocalNotifications({id: '1'})`.

What am I doing wrong?

southerneer:
Same problem for me. @superandrew213 's PR is a step in the right direction.

EdmundMai:
same issue here

EdmundMai:
Just fixed it with this. In the docs it says that you have to specify an object under the user `userInfo` options for iOS.

```
      PushNotification.localNotificationSchedule({
        id: 'myid',
        message: "Your event will begin in 10 minutes. Get ready!",
        userInfo: {
          id: 'myid'
        },
      });

      PushNotification.cancelLocalNotifications({
        id: 'myid'
      });

```

[...]

Issue #338, opened 2017-01-18
Android: Remove notifications from 'notification center'


JonesN3:
According to the documentation 'cancelLocalNotifications' is supposed to remove notifications from the 'notification center' in additiion to remove scheduled notifications by id. However, the dismissing of notifications does not seem to workinging on Android.

npomfret:
@varungupta85 we had this working a while back didn't we?

varungupta85:
@npomfret I have been using my own fork for the repo for some time as I have been having some problems with some of the commits and I didn't really need those features. But to answer your question, yes this has been working in the past and still works on my fork and I don't think that part of the code has changed, so it should be working now. There may be some problems with the way the API is invoked by @JonesN7 . @JonesN7, please keep in mind that you need to provide an object of shape `{id: <notificationId>}` to cancel a local notification by id and it will only cancel that notification. If you want to cancel all notifications, then you can use `cancelAllLocalNotifications` API. Please share your code if you are still having problems.

JonesN7:
Just to make sure we are on the same page; I am able to cancel local notifications by their ID, using `cancalLocalNotifications`. What this does not do for me, is remove a delivered notification from the status bar. So I can cancel future notification, but not remove delivered ones from the status bar, which from what I understand is the same as the 'notification centre' on iOS.

varungupta85:
@JonesN7 This is the line of code that should be called when a specific notification is canceled.

https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L424

`cancelScheduledNotification` is the method that should be called. I have a feeling that may be the notification is not canceled as well or you have more than one notification with different ids but same text. So, the other ones won't go away from the notification center. Can you confirm that you are seeing the log messages present in the `cancelScheduledNotification` method?

JonesN7:
I am trying to cancel two notifications. One that that has been delivered, and one scheduled but not delivered. The future one is correctly canceled, and this is shown the the log of the `cancelScheduledNotification` method. The one that has been delivered, and has an icon in the statusbar, does not show up in the log.

varungupta85:
The two notifications would have different ids unless they are repeating notifications. So, you will need to cancel both the notifications separately.

JonesN7:
[...]

Issue #259, opened 2016-11-01
Better documentation for cancelLocalNotifications() on Android


danleveille:
The documentation states that `userInfo` is iOS only, however, it states that you can use a `userInfo` dictionary as an argument for `cancelLocalNotifications()`.

Could someone whose more familiar with this library update the documentation to clarify:

- What does the argument for `cancelLocalNotifications` actually do (if anything) on Android?
- How do you cancel individual scheduled local notifications on Android? (What parameters can you use to match notifications?)

kenma9123:
Anyone can shed some light?

sticknor:
Hi, 
I was able to finally find success with this cancelLocalNotifications method. 


The README leads us to believe that the following will work for android:
PushNotification.localNotificationSchedule({ userInfo: { id: '123'} });
PushNotification.cancelLocalNotifications({ id: '123' });

But it does not....

Instead, do this:
PushNotification.localNotificationSchedule({ id: '123' });
PushNotification.cancelLocalNotifications({ id: '123' });

kenma9123:
I tried this `PushNotification.cancelLocalNotifications({ id: '123' });` but nothin works. Still can't remove a specific notification. So I end up doing the PR #336

j-nolan:
@sticknor 's solution did the trick for me (Android). Thanks!

I'd like to add that the key we pass through the `details` object **must** be called `id`:

```
// Does NOT work
PushNotification.localNotificationSchedule({ key: '123' }); // key, or anything else than id
PushNotification.cancelLocalNotifications({ key: '123' });

// Works
PushNotification.localNotificationSchedule({ id: '123' }); // id
PushNotification.cancelLocalNotifications({ id: '123' });
```

I think the documentation for this method should avoid the keyword `userInfo` because it suggests the matching will be performed using the `userInfo.id` property of existing notifications. Moreover, users can be mislead into believing any key can be used for matching. In reality only the `id` property can be used for matching.

If that makes sense I'd be happy to open a PR to change it. @npomfret

npomfret:
Yeah, lost of people find the readme a bit confusing - if you think you can improve please do

KatSick:
Confirm it working. But only with string type of property ID

GantMan:
Given I was in a situation where there were no give unique ids, and I didn't understand the 2 ways of cancelling for iOS and Android, I hope this helps others looking at this like:

[...]

PR #371, opened 2017-02-18
[Android] Fix cancelLocalNotifications

superandrew213:
Fixes [259](https://github.com/zo0r/react-native-push-notification/issues/259), [368](https://github.com/zo0r/react-native-push-notification/issues/368), [338](https://github.com/zo0r/react-native-push-notification/issues/338)

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -420,7 +420,7 @@ public class RNPushNotificationHelper {
                 String notificationAttributesJson = scheduledNotificationsPersistence.getString(id, null);
                 if (notificationAttributesJson != null) {
                     RNPushNotificationAttributes notificationAttributes = fromJson(notificationAttributesJson);
-                    if (notificationAttributes.matches(userInfo)) {
+                    if (notificationAttributes.getId().equals(userInfo.getString("id"))) {
                         cancelScheduledNotification(id);
                     }
                 }


southerneer:
Cheers, this is definitely an improvement. It would be nice to eventually have parity between iOS and Android in the canceling interface, but this works for now.

npomfret:
Does anyone know why the `matches` method isn't working?

superandrew213:
@npomfret because it was matching all object keys. We should only be matching ids since you are only supplying the id to cancel a notification.

npomfret:
I can't see this in the code, i looks right to me.  Its looping over the keys in the userInfo object passed in.  If they all match then we've found the notification to cancel - which is the same behaviour as iOS. No?

superandrew213:
@npomfret iOS uses the userInfo object to find a match, however userInfo is not available on Android. See [here](https://github.com/superandrew213/react-native-push-notification/blob/53b065340fdccbfebf36b317dfbc73982c5aa2c0/index.js#L122) and [here](https://github.com/superandrew213/react-native-push-notification/blob/53b065340fdccbfebf36b317dfbc73982c5aa2c0/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L90)

Is the cancelLocalNotification feature working for you with the current implementation on Android? I don't think it has ever worked.

The objects will never match since you are comparing all the notification attributes ```{ id: 123, message: "Hello", etc ...}``` with ```{id: 123}```.

npomfret:
This iOS UserInfo object is basically just a dictionary.  Below the iOS implementation of `cancelLocalNotifications`, I can't see where it's different to the java implementation we have here.  As far as I can see the algorithm is to loop through the keys of the passed in userInfo object, and check the values in each local notification object to see if they are all the same.  If they are all the same then remove it.


```
  for (UILocalNotification *notification in [UIApplication sharedApplication].scheduledLocalNotifications) {
    __block BOOL matchesAll = YES;
    NSDictionary<NSString *, id> *notificationInfo = notification.userInfo;
    // Note: we do this with a loop instead of just `isEqualToDictionary:`
    // because we only require that all specified userInfo values match the
    // notificationInfo values - notificationInfo may contain additional values
    // which we don't care about.
    [userInfo enumerateKeysAndObjectsUsingBlock:^(NSString *key, id obj, BOOL *stop) {
      if (![notificationInfo[key] isEqual:obj]) {
        matchesAll = NO;
        *stop = YES;
      }
    }];
    if (matchesAll) {
      [[UIApplication sharedApplication] cancelLocalNotification:notification];
    }
  }
```

superandrew213:
@npomfret there is nothing wrong with the match method. There is no userInfo object available on the Android implementation so how can you match userInfo objects? See [here](https://github.com/superandrew213/react-native-push-notification/blob/53b065340fdccbfebf36b317dfbc73982c5aa2c0/index.js#L122) and [here](https://github.com/superandrew213/react-native-push-notification/blob/53b065340fdccbfebf36b317dfbc73982c5aa2c0/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L90)

npomfret:
The userInfo object is just represented as a map in android (and iOS too - it's not a class, it's just a dictionary).  So from your js you call something like:

`cancelLocalNotifications({id: '1234'})`

Which get converted into a java `ReadableMap` object with one key and one value in Android, and a dictionary object in iOS.  From there the logic looks the same to me, so I don't understand why it isn't working.

Perhaps it's a type problem.  Are you passing a string for your id in when it should be an int, or vicer-versa?

superandrew213:
Ok so this {id: '1234'} ReadableMap object what does it get compared to?

npomfret:
I've got to suggest you take a look at the code, there's not much to it.

The _userInfo_ object get's passed in [here](https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L417).  And then it's compared to the each stored notification using the `RNPushNotificationAttributes` object which has the crucial `matches` method [here](https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L134).

Basically, each stored notification is treated like as a map.  Each key of the passed-in `userInfo` map is compared to the notification map... and if they all match - bingo.  Extra fields in the underlying notification **should** be ignored.

Have you tried stepping through in the Android Studio debugger to see if the match method is being called and why it's not matching your parameters?

superandrew213:
Ok so ```{id: '1234'}``` is being matched against each stored notification object? Right? So it will only cancel a notification if it finds a notification object that is equal to``` {id: '1234'}```.

However if it finds this notification object ```{id: '1234', message: 'Hello', etc..}``` then it won't cancel the notification since not all the keys match. That's the issue we have.

npomfret:
Extra keys in the underlying notification should not used in the comparison, same as with the iOS implementation.  And as far as I can see they are not.  See for yourself: https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L137

superandrew213:
Can you confirm that it works on your end and that you can cancel a notification? My PR is essentially just doing [this](https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L161). So I'm not sure why the match method doesn't work for me.

npomfret:
Aright, I think i understand why it isn't working.  The confusion is with the name - `cancelLocalNotifications` does not (and is not supposed to) remove notifications from the notification centre.  As I understand it the `cancelLocalNotifications` method only cancels **scheduled** notifications (on both iOS and android), ie. notifications that are persisted, and fire at some point in the future. And I believe it does work as intended.

The method that I think you want, which does not exist (yet).  And perhaps could/should be added as an android only method, something like:

```
    public void clearNotification(String id) {
        NotificationManager notificationManager = notificationManager();
        notificationManager.cancel(Integer.parseInt(id));
    }
```

This will remove an individual alert from the notification centre for android only.  I don't see why this can't be added so long as it's clear in the docs that it isn't supported in iOS.

superandrew213:
> As I understand it the cancelLocalNotifications method only cancels scheduled notifications (on both iOS and android), ie. notifications that are persisted, and fire at some point in the future.

This is what I am after. But it's ok. If it works on your end as it should then we can close this PR. I'll continue using my fork for now until I'll review this again when I'll upgrade.

npomfret:
Closing.  Lets add an android only _clearLocalNotification_ method

Shhzdmrz:
@npomfret are you going to add this `clearLocalNotification ` ? because `cancelLocalNotifications` never worked.

npomfret:
I'm afraid I don't have time right now.  You could submit a PR if you like?

PR #396, opened 2017-03-22
Implementing UNUserNotification functions

### LINKED ISSUE #389: [Feature Request] Use UNUserNotificationCenter in iOS 10
The new UNUserNotificationCenter has a method to remove the notification from the notification center...

Check this `https://developer.apple.com/reference/usernotifications/unusernotificationcenter/1649502-removealldeliverednotifications`

gperdomor:
Add functions to get and remove notifications from Notification Center

Fix #389

### DIFF
--- a/index.js
+++ b/index.js
@@ -306,4 +306,22 @@ Notifications.clearAllNotifications = function() {
 	return this.callNative('clearAllNotifications', arguments)
 }
 
+Notifications.removeAllDeliveredNotifications = function() {
+  if ( Platform.OS === 'ios' ) {
+    return this.callNative('removeAllDeliveredNotifications')
+  }
+}
+
+Notifications.removeDeliveredNotifications = function() {
+  if ( Platform.OS === 'ios' ) {
+    return this.callNative('removeDeliveredNotifications', arguments)
+  }
+}
+
+Notifications.getDeliveredNotifications = function() {
+  if ( Platform.OS === 'ios' ) {
+    return this.callNative('getDeliveredNotifications', arguments)
+  }
+}
+
 module.exports = Notifications;
--- a/README.md
+++ b/README.md
@@ -251,9 +251,27 @@ PushNotification.cancelLocalNotifications({id: '123'});
 
 `PushNotification.cancelAllLocalNotifications()` 
 
-Cancels all scheduled notifications AND clears the notifications alerts that are in the notification centre.
+Cancels all scheduled notifications.
 
-*NOTE: there is currently no api for removing specific notification alerts from the notification centre.*
+## Remove notifications
+
+### 1) getDeliveredNotifications (iOS only)
+
+`PushNotification.getDeliveredNotifications(callback: (notifications: [Object]) => void)` 
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -306,4 +306,22 @@ Notifications.clearAllNotifications = function() {
...
+  }
+}
+
+Notifications.getDeliveredNotifications = function() {
+  if ( Platform.OS === 'ios' ) {
+    return this.callNative('getDeliveredNotifications', arguments)
+  }
+}
+
 module.exports = Notifications;
```
npomfret: Need to add docs for all 3 methods.  Can we get android implementations too? If not the docs need to reflect this.

npomfret: ... also, where is the implementation of removeAllDeliveredNotifications, removeDeliveredNotifications & getDeliveredNotifications?

gperdomor: what do you mean with `where is the implementation`?

varungupta85: This package delegates the iOS work to the react native `PushNotificationIOS` package. I don't think that `PushNotificationIOS` has started using the new `UNUserNotification` library which is why @npomfret is asking for the implementation of these functions. Have you tested these methods using this package in an app?

gperdomor:
@varungupta85 @npomfret like i say in the PR comment, this depends on other PR https://github.com/facebook/react-native/pull/13036 which now is in process of integration, for that reason I put WIP in the title to prevent this was merged before the RN PR

index.js
gperdomor: @varungupta85 @npomfret like i say in the PR comment, this depends on other PR https://github.com/facebook/react-native/pull/13036 which now is in process of integration, for that reason I put WIP in the title to prevent this was merged before the RN PR... And yes, it was tested

varungupta85: @gperdomor Great stuff 👍  and thanks for the clarification. I missed your earlier comment in PR which created the confusion. Sorry about that. Looking forward to the react-native PR getting merged. I checked the react-native PR. Out of curiosity, how do I get the identifier for a notification to feed to the `removeDeliveredNotifications` that you added to `PushNotificationIOS` package.

gperdomor: @varungupta85 first you need the delivered notifications, then you need filter to get the notification or notifications you want to delete... Something like this:

```javascript
NotificationsIOS.getDeliveredNotifications((notifications) => {
    let toDelete = notifications.filter((n) => return YOUR FILTER CRITERIA);
    
    NotificationsIOS.removeDeliveredNotifications(toDelete)
});

varungupta85: Great, this would be really useful 👍

npomfret: This actually ties in nicely with a discussion elsewhere - android users wanting to remove specific delivered notifications from the notification centre.  Pretty sure it simple to achieve in android.

gperdomor: @varungupta85 @npomfret required code is now in react-native master branch

rrichardson: Is this waiting on anything in particular?  

Would it be too late to update this PR to handle `willPresentNotification`?

PR #397, opened 2017-03-22
[Android] handle mixed notification onNewIntent and getInitialNotification

carlobambo:
Broadcast onNotification to JS thread when receiving a mixed notifcation (with data and notification).
This will check if the intent has `notification` or `google.message_id` to know if intent is from GCM notification.
This will be useful for those who are using GCM on iOS and android like me.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -71,6 +71,7 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
       Bundle bundle = getNotificationBundle(intent);
         if (bundle != null) {
             bundle.putBoolean("foreground", false);
+            bundle.putBoolean("userInteraction", true);
             intent.putExtra("notification", bundle);
             mJsDelivery.notifyNotification(bundle);
         }
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -58,9 +58,18 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         return constants;
     }
 
+    private Bundle getNotificationBundle(Intent intent) {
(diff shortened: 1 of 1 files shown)


chromakode:
:+1: we had to make similar changes to get this working. https://github.com/zo0r/react-native-push-notification/compare/master...keybase:handle-android-mixed-notification-data?expand=1

jgab-net:
👍 I'm using your fork @carlobambo meanwhile @zo0r accept the PR. 👍

carlobambo:
@jgab-net just leave a comment here if you found any bugs. 😄

jgab-net:
Hi, @carlobambo, I did a [release](https://github.com/jgab-net/react-native-push-notification/releases/tag/2.2.1) in my fork with the stable commit (release 2.2.1) and your PR, because the master branch has a bug.

In your PR is missing a little change for handle notifications when the application its close, watch this [diff](https://github.com/jgab-net/react-native-push-notification/compare/stable-with-pr-397...jgab-net:fix-pr-397?expand=1), that happen when the config have this key `popInitialNotification: true`. 

@zo0r it would be nice that you add in the documentation a section explaining that is necessary `popInitialNotification: true` for handle notifications when the app its closed (not in background)

I did this [branch](https://github.com/jgab-net/react-native-push-notification/tree/feature-homologate-notifications) for homologate the notification object in javascript with iOS and manage the keys with json string in android bundles, 

@zo0r, @carlobambo, please watch this.

carlobambo:
@jgab-net, I've checked my commits and the line you said that is missing is there.
Btw what is bug you mentioned on master branch?

jgab-net:
@carlobambo Yes you're right, my bad.

After installing master of **zo0r** I get this error:
´JSON value '<null>' of type NSNull cannot be converted to NSNumber´

shrutic:
Not sure why this PR is not yet being accepted. @zo0r would really appreciate it if you could accept this PR because this is blocking from using mixed notification. Right now @carlobambo fork is behind the current master and I dont want to depend on it unless absolutely necessary as this would create one more level of dependency in our project

calcazar:
Working on adding this PR to my CE clone

Dallas62:
This PR is already implemented in codebase.

PR #404, opened 2017-03-30
Update docs for noisy remote push notifications

mdjnewman:
Thanks for the library, nice job wrapping PushNotificationIOS & adding a consistent Android option :)

This is a very small update - see https://firebase.google.com/docs/cloud-messaging/concept-options#messages-with-both-notification-and-data-payloads for details.


npomfret:
thanks for this

PR #421, opened 2017-04-27
Fix nested Bundle to JSON serialization

dmueller39:


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationJsDelivery.java
```diff
@@ -58,20 +58,29 @@ void notifyNotificationAction(Bundle bundle) {
     }
 
     String convertJSON(Bundle bundle) {
+        try {
+            JSONObject json = convertJSONObject(bundle);
+            return json.toString();
```
Komarev: nitpicking but maybe:
`return convertJSONObject(bundle).toString()`

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationJsDelivery.java
```diff
@@ -58,20 +58,29 @@ void notifyNotificationAction(Bundle bundle) {
...
-            try {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-                    json.put(key, JSONObject.wrap(bundle.get(key)));
-                } else {
-                    json.put(key, bundle.get(key));
-                }
-            } catch (JSONException e) {
-                return null;
+            Object value = bundle.get(key);
+            if (value instanceof Bundle) {
```
Komarev: Are you sure that in our case notification bundle has nested 'Bundle's and not just string (which looks like bundle)?

Komarev:
Looks good

npomfret:
Hi, thanks for this PR.  What's it for?  How does someone test it?  And what's the expected change in behaviour?

dmueller39:
This change is for notifications that contain nested dictionaries. You are able to create and send notifications that nest dictionaries, but when those same notifications are received, the nested dictionaries are dropped.

You can test it by checking out this short example: https://github.com/dmueller39/rnpnexample

If you use the example as is (which points to this branch) you should see the nested dictionaries:
screen shot 2017-04-29 at 6 00 02 pm

If you change package.json to point to the master version of the project, the nested dictionaries are dropped:
screen shot 2017-04-29 at 6 02 27 pm

dmueller39:
@npomfret ☝️

npomfret:
Many thanks.

PR #431, opened 2017-05-14
use _onNotification as handler in popInitialNotification if no handler is provided

blankg:
I believe the popInitialNotification is used mainly to control the timing of when to handle the initial notification (this is the case in my app) but the handling itself is no different than any other notification.
The proposed solution maintains backwards compatibility for passing in a custom handler but will use the _onNotification if no handler is passed. 
I think in most cases where popInitialNotification is used with current implementation people either duplicate the code in _onNotification or use the private _onNotification as the handler for popInitialNotification.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["react-native"]
+}
--- a/README.md
+++ b/README.md
@@ -212,6 +212,13 @@ PushNotification.localNotification({
 });
 ```
 
+## Handling Initial Notification
+By default the initial notification is popped automatically.
+
+To manually pop the initial notification configure the `popInitialNotification` value to `false` and call `PushNotification.popInitialNotification()`. This will result in your `onNotification` handler being called with the initial notification.
+
+Optionally pass a custom handler function to the `PushNotification.popInitialNotification` method if you would like to have a different handler for initial notifications than the `onNotification` handler.
+
 ## Scheduled Notifications
 `PushNotification.localNotificationSchedule(details: Object)`
 
--- a/__tests__/index.test.js
+++ b/__tests__/index.test.js
@@ -1,0 +1,100 @@
+'use strict';
+
+import 'react-native';
+
+describe('Test popInitialNotification method', function() {
+    let Index;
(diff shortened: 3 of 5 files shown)


npomfret:
hi - thanks for this pr!

Can you detail:

 * what is the expected change in behaviour
 * how would someone else test it

And if it's an API change it will need to be documented.

blankg:
Hi @npomfret, thanks for quick response.

The popInitialNotification method is not currently documented (just the configuration is documented), should I add it to the documentation?

As for the requested details:

- what is the expected change in behavior?

The PR keeps backwards compatibility, so for people that pass in a handler function as before it will behave the same. The change is that now you can also use the popInitialNotification without passing a handler and it will use the _onNotification as the handler.

- how would someone else test it?

**Test1**: call popInitialNotification with a custom handler function.
Test configuration: popInitialNotification: false.
Test flow: kill app -> open app from a notification.
Expected result: custom handler should receive result object as before (the result from callNative('getInitialNotification')).

**Test2**: call popInitialNotification with no parameters.
Test configuration: { popInitialNotification: false, onNotification: onNotificationHandler}
Test flow: kill app -> open app from a notification.
Expected result: onNotificationHandler should be called with a notification object.

**Test3**: No explicit call to popInitialNotification, configure popInitialNotification to true.
Test configuration: { popInitialNotification: true, onNotification: onNotificationHandler}
Test flow: kill app -> open app from a notification.
Expected result: onNotificationHandler should be called with a notification object.

npomfret:
Thanks for the info. I'm still a bit swamped and finding it difficult to get the time to test these PR>  Could you ask someone else who has an outstanding PR to test yours (and vice versa)?

blankg:
@npomfret , Tried to react out to do test swapping but didn't get a response. So I have added jest tests to the popInitialNotification which unit tests the above three test scenarios.
You can run these test with "npm test" (after running "npm install" in order to get jest).
The jest integration can be very useful for testing changes in the js code (less helpful if changes are the native side).

npomfret:
@blankg thanks for trying at least.  I'll see if I can test it at some point.

With regard to the test plan...

What notifications need to be sent to the device, and for each test, when should the notifications be sent? (Maybe before the test starts?)

Should the notifications be 'noisy' or 'silent'?

At the start of each test, what state should the app be in (killed / background / foreground)?

We are trying to achieve consistent behaviour between iOS and Android.  Does the change in behaviour caused by the PR improve the consistency or reduce it?

blankg:
Hi @npomfret ,
This PR keeps behavior consistent between iOS & Android (also there are no native code changes here).
As for the test plan:
- For the unit tests I've added everything is mocked so you just need to run them.
- For manual testing the app should be killed, then send a loud push and open the app from the notification.
Thanks again!

PR #468, opened 2017-06-30
[Android] Added ability to specify `largeIcon` as URL.

tschoffelen:
Not only can you now specify `largeIcon` as a bundle item, but when the value starts with `http`, it gets downloaded, rounded, and set as the large icon of the notification.


zorosun:
@tschoffelen 

Does it work to use `file://...` path as URL ?

tschoffelen:
Not with this version. Either `http://` or `https://`, or as a bundle string that resolves to the image path. 

Allowing `file://`would require changing [line 150](https://github.com/zo0r/react-native-push-notification/pull/468/commits/db18a3bade867420248477850ffd5b11be76d7d2#diff-daa707ee59f959fbde5b91695d70012eR150) to check for something like `://` instead.

zorosun:
It works! thank you.

Angelk90:
@tschoffelen: How do I place a Large notification icon, an image taken from an http url.
Example: https://batch-doc.s3.amazonaws.com/Push/BigIcon.png
Is not it very clear to me? I need that the notifications come from the server.

tschoffelen:
That is exactly what this PR makes possible. Simply add a `largeIcon` key to your notification payload (where you also specify the notification text) with a string value for the string of the URL of the image.

Angelk90:
@tschoffelen: So I have to change the original file with this: https://raw.githubusercontent.com/tschoffelen/react-native-push-notification/db18a3bade867420248477850ffd5b11be76d7d2/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java

tschoffelen:
Yep! That should do it.

Angelk90:
Does not work:
Server parse:
```
Parse.Push.send({
    channels: channel,
    data: {
      title: title,
      alert: text,
      objectId: postId,
      badge: "Increment",
      sound: "default",
      color:"#006ACC",
      largeIcon: "https://images.vanityfair.it/gallery/58873/Mid/248b4ee8-d0e8-4e0e-8dcc-882ca863f536.jpg"
    }
  },{
    useMasterKey: true,
    success: function(){
      console.log("We pushed!!");
    },
    error: function(error){
      console.log("Some error occurred:"+error);
    }
  });
```

I receive as notification:
immagine

Where am I wrong?

tschoffelen:
Your server-side code seems fine. Did you remember to recompile after changing the file?

Angelk90:
You do, you can see me: console.log ('NOTIFICATION:', notification);
What prints in detail.

Angelk90:
And a screen of how it should be the notification I receive.

tschoffelen:
That does not mean the Java code has been recompiled, the javascript just outputs whatever you put in the payload. The result should be exactly the way your screenshot showed, we use it in our company for that exact purpose.

Angelk90:
@tschoffelen: How do i recompiled the java code?

Angelk90:
I took the file it replaced and then from cmd i invoked: react-native run-android

zorosun:
@Angelk90  are you using localNotification to push?

Angelk90:
@zorosun: No, they are notifications that come from a server, so:
           OnNotification: function (notification) {
               Console.log ('NOTIFICATION:', notification);

zorosun:
this is what I did

`
onNotification: async function(notification) {
      console.log('NOTIFICATION:', notification);
      if (notification.userInteraction) {
          // when use click notification from home screen
      } else {
          // onReceive from server when push notification
          // I pushed to localNotification to get what I want

              bigText: 'message',
              subText: 'message',
              largeIcon: 'http://image_url_from_server',
              title: 'title',
              message: 'message'
           });
      }
    },
`

Angelk90:
@zorosun: But when you are out of the application do not receive the double notification?

zorosun:
@Angelk90 I didnt do the server but when OnNotification called my phone doesnt show anything.

Angelk90:
@zorosun: Instead, my device shows two notifications when I'm out of the app.
@tschoffelen: How do I compile a module?

zorosun:
react-native run-android should recompile

Angelk90:
@zorosun: I tried with: react-native run-android should recompile
Nothing does not work either, you always see the image of the ic_launcher.

zorosun:
not sure if you replace the file correctly but you can try 

- replace master branch with @tschoffelen's one in package.json 
- remove node_modules/react-native-push-notification
- npm install again
- react-native run-android

if it doesnt work then I have no idea

Angelk90:
@zorosun: I took the file: https://raw.githubusercontent.com/tschoffelen/react-native-push-notification/db18a3bade867420248477850ffd5b11be76d7d2/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
Replaced it with the file: node_modules \ react-native-push-notification \ android \ src \ main \ java \ com \ dieam \ reactnativepushnotification \ modules \ RNPushNotificationHelper.java
What I had, and I compiled as you told me.
but it does not work.

Angelk90:
@zorosun: Do you think I'm wrong?
@tschoffelen: Where am I wrong?

tschoffelen:
I think there might be some cache files (possibly in `android/build` or `android/app/build`) that still have the old version - you might try removing (or moving just to be safe) those directories and re-run the app. Also note that I think you have to fully stop the running app after the new version is installed, and then start it again to register the new notification handling code. 

If that doesn't work, look at if the RNPushNotificationHelper.java is also somewhere else (globally installed or copied by React Native), you might have patched the wrong version of the file. Telling NPM to get the package from my repo would be the easiest way to resolve this as @zorosun suggested.

I just checked - this is the exact version of the code we use in production, so the code itself definitely works.

tschoffelen:
Sorry to the reviewers for this tangent by the way, didn't mean to mess up the PR with troubleshooting...

Angelk90:
@tschoffelen: I've deleted app from the device.
Canceled module, and added to your:
https://github.com/tschoffelen/react-native-push-notification
Then I did: cd android && gradlew clean
But it does not work.

You could tell me what to print: Console.log ('NOTIFICATION:', notification);
Is OnNotification Empty?

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -229,30 +295,6 @@ public void sendToNotificationCentre(Bundle bundle) {
...
-                            soundName = soundName.substring(0, soundName.lastIndexOf('.'));
-                            resId = context.getResources().getIdentifier(soundName, "raw", context.getPackageName());
-                        }
-
-                        soundUri = Uri.parse("android.resource://" + context.getPackageName() + "/" + resId);
-                    }
-                }
-                notification.setSound(soundUri);
-            }
-
```
godness84: @tschoffelen why you deleted these lines?

godness84: I meant lines 232-255

tschoffelen: No idea, think this must be added later or something. Added it back in.

godness84:
Have you seen my PR that is based on yours? It adds a "largeIconCircular"
key, but mostly it fixes an issue with the rounded image when it's
rectangular and not perfectly squared.

vvavepacket:
when is this feature going into master so that we can use https images on largeIcon ?

vvavepacket:
@godness84 

which source code of yours contains 'largeIconCircular', because I pulled latest fork of tschoffelen  and can't find largeIconCircular anywhere. I tried putting largeIconCircular as false, but it is still rounded icon

wesleymc:
@tschoffelen Have you already tried receiving a remote notification with the application terminated? It seems to me that the approach of your PR crash the application when it is closed and gets a notification with a largeIcon URL to deal with. Error.

sauhardar:
Anyone know how to position the large icon (as passed by URL) to the right of the title/bodyText or how to make it round? Is there any way of styling icons received through URL?

PR #500, opened 2017-08-02
Remove depreciated createJSModules marker - RN 0.47 compatibility

radko93:
[`createJSModules` is now not required on Android](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8) from RN 0.47. I actually can't build an app on RN 0.47 without removing this method/ovveride marker. This is backwards compatible according to my tests.

PR #505, opened 2017-08-05
Fix Android compilation error with RN 0.47

dluksza:
Method createJSModules was removed from RN 0.47. Which leads to compilation error due to @Override annotation. This method can be removed right now.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/ReactNativePushNotificationPackage.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/ReactNativePushNotificationPackage.java
@@ -18,11 +18,6 @@ public class ReactNativePushNotificationPackage implements ReactPackage {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


Parrryy:
Can confirm this allowed me to compile in 47.

tommeier:
Only removing the `@override` maintains backward compatibility

valeryq:
In accordance with other libs.
```
  // Deprecated RN 0.47
  public List<Class<? extends JavaScriptModule>> createJSModules() {
    return Collections.emptyList();
  }
```

vbuch:
@dluksza could you change your PR so that it only removes the `@Override`? This way it remains BC as @tommeier stated above. Many others do it the way @valeryq showed. Then @zo0r can safely merge it in.

maxkomarychev:
@vbuch there's another PR opened couple of days prior to this which is backward compatible - #500

PR #508, opened 2017-08-08
Remove deprecated createJSModules

ivanzotov:
createJSModules deprecated in RN 0.47

PR #518, opened 2017-08-10
Remove createJSModules @Override

arv:
createJSModule was removed in React Native 0.47. By removing the override annotation this now compiles correctly again in all versions of React Native.


arv:
Closing in favor of #508

PR #549, opened 2017-09-05
Update README.md

lucask42:
Added info to help with a gotcha we encountered whilst using this package.  Lacking the super call meant that the extras from the notification were ignored.

We couldn't get onNotifications to fire on Android local scheduled notifications and we found that during an installation of another module, namely React Native Branch Deep Linking, onNewIntent was overridden without adding the super.onNewIntent() bit.  This and possibly other RN modules could be leading to some of the difficulties with onNotification not firing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,17 @@ public class MainApplication extends Application implements ReactApplication {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Dallas62:
Hi @lucask42 
This PR is old, but do you know if it's still true ?

flochtililoch:
Still relevant today, even though Branch now seem to have [updated their doc](https://help.branch.io/developers-hub/docs/react-native#android) to include the super call

PR #554, opened 2017-09-12
add "inboxStyle" property to notification (Android only)

godness84:
# Motivation
The library is missing the possibility to set the InboxStyle to the notification. This PR adds it.
For details about the InboxStyle: https://developer.android.com/reference/android/app/Notification.InboxStyle.html

# Usage
`PushNotification.localNotification()` and `PushNotification.localNotificationSchedule()` now accepts a new optional property that is `inboxStyle`. Following an example:

```
      inboxStyle: {
        bigContentTitle: "Topic",
        summaryText: "3 messages",
        lines: [
           "John: this is just an example",
           "Brad: of course!",
           "last line, very very very very long that will not wrap but will be truncated by the inbox style"
        ]
      }
    });
```

# Test plan
I added the `inboxStyle` property to the object passed to `PushNotification.localNotification()` and `PushNotification.localNotificationSchedule()` and the notification gets displayed correctly.
If I don't add the `inboxStyle`, the behavior does not change.

# About the code
I needed to copy-paste the code of `Arguments.java` from react-native repo because, before RN 0.48.0, `Arguments` was not able to convert lists from ReadableMap to Bundle. I wanted to make this PR compatible with older versions of RN, so I copied the code into the `helpers` namespace.

I also needed to add a `RNUtil` class in order to convert a `json` to `WritableMap` and vice versa.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,14 +211,30 @@ PushNotification.localNotification({
     number: '10', // (optional) Valid 32 bit integer specified as string. default: none (Cannot be zero)
     repeatType: 'day', // (Android only) Repeating interval. Could be one of `week`, `day`, `hour`, `minute, `time`. If specified as time, it should be accompanied by one more parameter 'repeatTime` which should the number of milliseconds between each interval
     actions: '["Yes", "No"]',  // (Android only) See the doc for notification actions to know more
-	inboxStyle: { // (Android only) See the android doc (https://developer.android.com/reference/android/app/Notification.InboxStyle.html)
-       bigContentTitle: 'title', // (optional)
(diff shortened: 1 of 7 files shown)


tarunteckedge:
Any idea by when this can be merged?

Thanks in advance!

godness84:
@tarunteckedge github bot replied 🤣 
it's a pity that it didn't get merged

Dallas62:
Hi @godness84 
I will look into it ASAP.

Dallas62:
Hi @godness84 
I think we should look into another way to implement this kind of Styling.
I try to close look into when I got time.

PR #557, opened 2017-09-14
Feature/twilio push

marneborn:
If twi_body is set on the message, set that as the "message" field for the android notification bar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,11 @@ PushNotification.configure({
     // (required) Called when a remote or local notification is opened or received
     onNotification: function(notification) {
         console.log( 'NOTIFICATION:', notification );
+
+        // process the notification
+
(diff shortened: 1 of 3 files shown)


evancorl:
@marneborn @philberg Hey, just wondering why you decided to close this PR and delete the branch? My team and I are actually having the same issue with Twilio and were fixing to make a similar PR. Maybe a version that allows the user to specify the name of the `message` field.

marneborn:
We are using a fork of react-native-push-notification in our project, I
meant to make the PR against that...
This way of fixing it is too specific, I think, for a library.
I would support a more generic solution, but couldn't think of a
good/robust one. Mostly because I just started looking at this stuff and
don't know the cases I would want to support.

-mikael

evancorl:
@marneborn I see. We'd prefer not to fork the repo based on something so minor. We will probably create an issue and work on a PR in the meantime. Allowing the user to specify the name of the `message` field feels like the most generic solution.

PR #574, opened 2017-10-14
Make local notifications on android notify the application too

ErikMikkelson:
RNPushNotificationPublisher receives the local notifications, but it doesn't route them to the application if it's in the foreground like the ListenerService does.  This makes it work the same for both, and thus duplicates the behavior that we see on iOS.

Also, I missed it when I first set up the code, but it appeared that userInteraction was overridden to false in the code, but I removed that so we can know if we were clicked from the notification center.

Hey, for people trying to handle background notifications on Android, the intent filter trick listed in [122](https://github.com/zo0r/react-native-push-notification/issues/122) worked for me.

I can now receive local notifications in the foreground and background, warm and cold start on iOS and Android and onNotification is called.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -97,7 +97,6 @@ public class RNPushNotificationListenerService extends GcmListenerService {
 
         RNPushNotificationJsDelivery jsDelivery = new RNPushNotificationJsDelivery(context);
         bundle.putBoolean("foreground", isForeground);
-        bundle.putBoolean("userInteraction", false);
         jsDelivery.notifyNotification(bundle);
 
         // If contentAvailable is set to true, then send out a remote fetch event
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -56,7 +56,7 @@ public class RNPushNotificationListenerService extends GcmListenerService {
         handler.post(new Runnable() {
             public void run() {
                 // Construct and load our normal React JS code bundle
-                ReactInstanceManager mReactInstanceManager = ((ReactApplication) getApplication()).getReactNativeHost().getReactInstanceManager();
+                final ReactInstanceManager mReactInstanceManager = ((ReactApplication) getApplication()).getReactNativeHost().getReactInstanceManager();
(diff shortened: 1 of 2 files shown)


ital0:
Please @zo0r accept this PR :)

nonameolsson:
@zo0r Have you had time to try it? Looks good!

ErikMikkelson:
My team has been using this fork for the last three weeks and so far no problems.

nonameolsson:
Could it be merged? It's really important to get this one working.

amazingmarvin:
I have tested local notifications when app is in the foreground with this PR and it works just like with `PushNotificationIOS`.

Thanks @ErikMikkelson!

supercamilo:
Merge plz

royipressburger:
Will be happy to see that as an official solution to the problem, and not using other libraries like react-native-fcm

ljuborados:
This pull request doesn't work with ExpoKit,
failing on notification received with:
```
FATAL EXCEPTION: main ... java.lang.ClassCastException: com.mybundle.MainApplication cannot be cast to com.facebook.react.ReactApplication
```

calcazar:
Added this PR to the CE clone

buddhamangler:
This pull request saved my butt.  Please merge

andrzejbe:
a PR with plenty of positive comments - that has no conflicts, adds long-awaited functionality (which makes local notifications finally usable) and works like a charm.. Please merge!

bgerhards:
This seemed to fix the issue, but now it's old enough that Android 8 notification requirements are missing. This requirement regarding the mandatory notification channel has been resolved in enhancements already added to the package. Update your repo @ErikMikkelson and keep me up to date. I hope @zo0r approves this PR, it solved my recent issues with foreground notifications.

ErikMikkelson:
It's 2020 and the changes listed above still aren't in.  @zo0r @npomfret @bgerhards I just pulled from master and compared old changes with new files and they appear the same.  I'm not sure what changes are necessary for Android 8?  Most of my changes were around getting RNPushNotificationPublisher to work the same as RNPushNotificationListenerService while getting the application context a different way.

Dallas62:
This PR has been implemented on `dev`.

Dallas62:
Hi,

I merged this PR, but it seems Apple has deprecated this feature:
https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622930-application?changes=latest_minor&language=objc

And alternative isn't handling background notification when there is no user actions...
https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate

I'm not sure allowing this on Android is a good idea anymore since there is no way to do this on iOS.

What do you think ?

PR #581, opened 2017-10-22
Add support attachment-url on Android

kmcrawford:
This adds support for attachment-url for Android like iOS supports today.


Dallas62:
Hi !
This is available on `dev` branch.

PR #638, opened 2018-01-22
Fix id in present/scheduleLocalNotification.

amazingmarvin:
In the docs it says you can provide a notificationId in `scheduleLocalNotification` like this:

```javascript
PushNotification.scheduleLocalNotification({
  id: Math.floor(2147483648*Math.random()),
  ...
});
```

but this doesn't work and a random ID is always generated by `react-native-push-notification` since it's read from the arguments bundle incorrectly.

I don't think this is a breaking change since trying to specify an `id` as a string causes a crash anyway.


amazingmarvin:
According to #639, string IDs are working.  Closing for now until I can investigate more.

Issue #645, opened 2018-01-29
Consider updating ShortCutBadger for Android


yshing:
By updating the ShortCutBadger from 1.1.8 to 1.1.21 ( current latest)
Should fix various bugs on different devices.
e.g. The `tryLegacySamsungBadge` and code [here](https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/helpers/ApplicationBadgeHelper.java) related to issue 40 in ShortCutBadger can be safely removed.

And would not throw Error for Sony machines when receiving Push with badge count from server side.

PR #646, opened 2018-01-29
Updating ShortCutBadger for Android

yshing:
#645


tahajahangir:
@zo0r 
ShortcutBadger version is stuck to 1.1.8, but latest version is 1.1.23.
No plan to upgrade it?

PR #623, opened 2018-01-09
[Android] Adding support for big image notifications

shivam-aditya:
I have made changes to support big image notifications for the library. Also made changes to the readme reflecting the same.

PR #668, opened 2018-03-08
[Android] Big Image support for android push notification.

nodexpertsdev:


nodexpertsdev:
This is exactly same as Pull request #623 with just some changes made with respect to the updated status of the library. Added project dependency in addition to #623 without dependency it will not throw the error but the image will not be loaded.

saifkamaal:
is this only for android or iOS also ?

leventyildirim61:
Hello @saifkamaal 

I think that you should use rich notification for iOS. Here is example https://www.pluralsight.com/guides/creating-ios-rich-push-notifications.

Good luck.

IgorMitev:
Thank you for the feature, is there a way to configure the image to be shown in contain mode rather than cover (avoid being cropped)?

Gp2mv3:
I corrected the conflict. Can someone test this ?

Neogohan1:
Thanks for that, tested in mine and seems to work great with some slight changes to resize the image to fit a bit better.

Dallas62:
I fixed the conflict but not yet merging this.
The fresco dependencies is not up to date anymore, I will check it ASAP.
I'm a bit confuse there is no fallback for iOS without implement custom code, do you have information on this ?
Regards,

IgorMitev:
@Dallas62 Can you please let me know if there is a plan for the feature to be merged at some point or it's abandoned?

Thank you!

Dallas62:
@IgorMitev https://github.com/zo0r/react-native-push-notification/blob/dev/CHANGELOG.md

PR #684, opened 2018-03-21
Add gradle configuration property

superandrew213:
Fixes https://github.com/zo0r/react-native-push-notification/issues/682

PR #686, opened 2018-03-22
Compilation error

hirthbrian:
Hello,

I don't know if it's a problem for everyone but there is a compilation error on Android with the updated Google Play services.

Prevent Error: more than one library with package name 'com.google.android.gms.license' with Google Play services 12.0.0


cosmith:
This PR already has a better way of solving this: https://github.com/zo0r/react-native-push-notification/pull/684

hirthbrian:
Awesome, thanks!

PR #687, opened 2018-03-23
limit the components that Intent will resolve to

### LINKED ISSUE #648: [Security issue] Leakage of all pushes to third-party apps
https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L311
https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationRegistrationService.java#L36

All your users are vulnerable to this bug. I submitted an email to you a few months ago, but no fix/response was received. 

You don't specify broadcast receiver in the broadcasts above. It means that any app installed on the same device can intercept those broadcasts, so all pushes from server are leaked to third parties. Use methods like ``` Intent.setPackage() ``` or ``` Intent.setClass() ``` etc to prevent broadcast interception

vstostap:
fix #648


Gp2mv3:
Need to test that but it seems to be ok for me.

PR #690, opened 2018-03-26
Add gradle configuration property

### LINKED ISSUE #682: Compatibility with react-native-firebase `Error: more than one library with package name 'com.google.android.gms.license'`
Hello,

I'm trying to use this library along with `react-native-firebase` in my project.
But when I try to compile, I'm getting the following error:

```
:react-native-push-notification:processReleaseResources FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-push-notification:processReleaseResources'.
> Error: more than one library with package name 'com.google.android.gms.license'

```

I believe your Android dependency  `compile 'com.google.android.gms:play-services-gcm:+'` needs to support Gradle versioning, kinda like: 
https://github.com/rebeccahughes/react-native-device-info/blob/e2c877f461c4181113abb3ff8f6bb8dfe89de0fb/android/build.gradle#L30
so the clients have the option of explicitly setting a version for that dependency.

Other packages seem to be affected as well by this change:
https://github.com/rebeccahughes/react-native-device-info/issues/365

lfkwtz:
Update of PR #684 to fix #682

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,6 @@ The component uses PushNotificationIOS for the iOS part.
 
 In your `android/build.gradle`
 ```gradle
-...
-
 ext {
     googlePlayServicesVersion = "<Your play services version>" // default: "+"
 }
--- a/README.md
+++ b/README.md
@@ -36,17 +36,14 @@ The component uses PushNotificationIOS for the iOS part.
 
 ## Android manual Installation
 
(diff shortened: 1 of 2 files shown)


Macarse:
Thanks for the fix!
When should I expect 3.0.3 to be released?

PR #694, opened 2018-03-28
Android: Play default sound if file name at "soundName" does not exist in the "raw" directory

Grytsiv:
Android: Play default sound if file name at "soundName" does not exist in the ''raw" directory. Also fixed crash when wrong file name at "soundName".


Gp2mv3:
Interesting PR ! 
I'll check it later (in the example app) to be sure it's working.

In the meanwhile, can you add a note in the Readme ?

Thanks,

Gp2mv3

Issue #410, opened 2017-04-05
Remote silent PN fires when app in bg/closed mode


igor-bv:
Guys, please, help me :)
When app is closed or in bg mode and when I receive PN with flag `silent === 'false'` I got notification in status bar with sound as I expecting, but when I receive PN with flag `silent === 'true'` I dont want to have any sound/alert/anything for user, but notification appears with default title and text...
Two weeks later(approx) such issue didn't appear.
Can anyone help or suggest with something?
Thanks in advance!

PR #707, opened 2018-04-10
[Android] Remove default alert for silent push notifications

superandrew213:
On Android silent push notifications trigger alert with `Notification received` message when app is in background.

If there is no `message` supplied then user should not be alerted with a default message. Otherwise it makes silent push notifications impossible on Android.

At the moment every time a silent push notification is received (while app is in background) user receives an alert with `Notification received` message.

Fixes: https://github.com/zo0r/react-native-push-notification/issues/410

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
@@ -29,7 +29,7 @@ public class RNPushNotificationListenerService extends GcmListenerService {
         JSONObject data = getPushData(bundle.getString("data"));
         if (data != null) {
             if (!bundle.containsKey("message")) {
-                bundle.putString("message", data.optString("alert", "Notification received"));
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationListenerService.java
```diff
@@ -29,7 +29,7 @@ public void onMessageReceived(String from, final Bundle bundle) {
         JSONObject data = getPushData(bundle.getString("data"));
         if (data != null) {
             if (!bundle.containsKey("message")) {
-                bundle.putString("message", data.optString("alert", "Notification received"));
+                bundle.putString("message", data.optString("alert", null));
```
superandrew213: @zo0r is there a reason why `alert` is used as the field and not `message`?

superandrew213:
@zo0r & @npomfret can you have a quick look at this. It's a minor change that can be quickly merged. It shouldn't break anything, unless there is a reason why it was done like this.

PR #711, opened 2018-04-12
Add clearLocalNotification method

FlorianGerhardt:
As discussed in #371 there is no way to remove a notification from the notification centre.
This PR adds a method that removes a notification by id from the centre.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -195,6 +195,14 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
     }
 
     @ReactMethod
+        /**
+         * Clear notification from the notification centre.
+         */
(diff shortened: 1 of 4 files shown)


TheBerg:
This is great, I was just looking into this. 

We should probably also support removing via a tag too. 

```/**
     * Cancel a previously shown notification.
     * @param tag the string identifier of the notification.
     * @param id the ID of the notification
     */
    public void cancel(@Nullable String tag, int id)```

Let me know if you need help. :)

**Comment by rstiyer** [comment]
Could you please give an example of how to use this function?

**Comment by joshbax** [comment]
This method only works on Android. 
React-native is intended to be a cross platform framework. If you can't avoid adding platform specific functions, at the very least they should be documented or appropriately named (e.g. `clearLocalNotificationAndroid()`)

**Comment by noway** [comment]
agree with @rstiyer how to use it?

**Comment by noway** [comment]
this is how:
```
PushNotification.clearLocalNotification(parseInt(notificationToRemove.androidNotificationId, 10));
```

PR #758, opened 2018-06-08
Hard-code a notification channel to use on Android

akalin-keybase:
This is required on Android Oreo.


akalin-keybase:
Sorry, wrong repo.

PR #772, opened 2018-06-27
Update README.md

eridem:
Add dependency for the android/build.gradle


Gp2mv3:
Why do you need this dependency here ?
Dependencies are in settings.gradle no ?

PR #786, opened 2018-07-11
Update build.gradle

bhram:
solve this issuse
https://stackoverflow.com/questions/45462666/notificationcompat-builder-deprecated-in-android-o


Gp2mv3:
Isn't it a regression ?
As you can configure the values of the gradle config as explained in the readme. Your solution hardcodes the version 26, which is less polyvalent.

Gp2mv3:
I'll close this PR to clean a bit the repo, feel free to open it back if you think it was closed by mistake.

Issue #547, opened 2017-09-05
PushNotification.popInitialNotification returns null


csooraj:
On clicking notification I am able to get call inside
PushNotification.popInitialNotification(notification) but the notification data is null, I need to route user to different screens based on the notification data.

Could any one please help.

Here is my function, the notification field is printed as null.

PushNotification.popInitialNotification(notification => this.handleNotification(notification));

handleNotification(notification) {
console.log("The notification is----->", notification);
}

Clcll:
@csooraj Hi! did you solve the issue ? my popInitialNotification function works only once

ComicScrip:
Same  problem here, on Android, popInitialNotification handler always returns null when app is in background or killed. Everything is working properly on iOS. 

I checked out the installation steps a thousand times, I think my setup is fine...
(Using RN 0.45.1 and react-native-push-notification 3.0.1)

@csooraj did you manage to solve it ?

ComicScrip:
Finally solved the sh*t out of this ! 
I know it's probalbly very hacky but it works. Modify the content of node_modules/react-native-push-notification/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java as follow : 

```java
package com.dieam.reactnativepushnotification.modules;

import android.app.Activity;
import android.app.Application;
import android.app.NotificationManager;
import android.content.BroadcastReceiver;
import android.content.Context;
import android.content.Intent;
import android.content.IntentFilter;
import android.os.Bundle;

import com.dieam.reactnativepushnotification.helpers.ApplicationBadgeHelper;
import com.facebook.react.bridge.ActivityEventListener;
import com.facebook.react.bridge.Arguments;
import com.facebook.react.bridge.Promise;
import com.facebook.react.bridge.ReactApplicationContext;
import com.facebook.react.bridge.ReactContext;
import com.facebook.react.bridge.ReactContextBaseJavaModule;
import com.facebook.react.bridge.ReactMethod;
import com.facebook.react.bridge.ReadableArray;
import com.facebook.react.bridge.ReadableMap;
import com.facebook.react.bridge.WritableMap;

import java.util.HashMap;
import java.util.Map;
import java.util.Random;

public class RNPushNotification extends ReactContextBaseJavaModule implements ActivityEventListener {
    public static final String LOG_TAG = "RNPushNotification";// all logging should use this tag

    private RNPushNotificationHelper mRNPushNotificationHelper;
[...]

PR #793, opened 2018-07-16
Update RNPushNotification.java for notification retrieval when the app is in the background/killed

akbokha:
As reflected in https://github.com/zo0r/react-native-push-notification/issues/547 - ComicScrip's suggestion.

_Rationale:_ While local notifications work properly in the following three use-cases: app in the foreground, background and killed, this is not the case for remote notifications.  For remote notifications, only the first use case (foreground), results in a notification object one can work with. The other two use cases do work properly, but one does not retrieve a notification object in the popInitialNotification callback (i.e. it is null). 

The _expected behaviour_ of the application w/ these changes is thus:
- One is able to receive a remote notification (and being able to retrieve its respective object) while the app is in the background
- One is able to receive a remote notification (and being able to retrieve its respective object) while the app is killed.
- All the other functionalities (e.g. receiving a remote notification while the app is in the foreground, local notifications work properly in all the three use cases etc) remain functional

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -8,6 +8,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.os.Bundle;
+import android.support.v4.app.NotificationManagerCompat;
 
 import com.dieam.reactnativepushnotification.helpers.ApplicationBadgeHelper;
 import com.facebook.react.bridge.ActivityEventListener;
@@ -116,6 +117,13 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
     }
 
     @ReactMethod
+    public void checkPermissions(Promise promise) {
+        ReactContext reactContext = getReactApplicationContext();
+        NotificationManagerCompat managerCompat = NotificationManagerCompat.from(reactContext);
+        promise.resolve(managerCompat.areNotificationsEnabled());
+    }
+
+    @ReactMethod
     public void requestPermissions(String senderID) {
         ReactContext reactContext = getReactApplicationContext();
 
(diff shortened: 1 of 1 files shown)


voxuanthuan:
I got error
RNPushNotificationHelper.clearNotification(notificationID);
                                 ^
  symbol:   method clearNotification(int)
  location: variable mRNPushNotificationHelper of type RNPushNotificationHelper
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.

akbokha:
@voxuanthuan Yeah, but note that this error is not part of this pull request.

While the statement in question is actually part of the diff, it is not something that I functionally changed (see master). I only changed the spacing since it was formatted properly.

The fix for this error is straightforward, but someone should make a separate PR for this.

Gp2mv3:
I've to test this PR with a real app and then we can merge it.
Do you have an example app or a test that can demonstrate its behavior before merging ?

Gp2mv3:
Can you test with the new example app I pushed some minutes ago ? In the example folder of master.

akbokha:
Sorry for the late reply @Gp2mv3. Is it okay if I get back to you this weekend/the end of the week?

Short update: We are currently using a [fork]{https://github.com/datacamp/react-native-push-notification} of the react-native-push-notification lib and actually are using it in production right now (can confirm that it works only with these changes), but the master branch of this repo seems to be relatively unstable. Which is why we went back to 3.0.2 (the version we were using) and applied the changes reflected in the diff of this PR there. Everything worked smoothly from that point onwards.

Will get back  to you on the tests performed with the example app by the end of this (work)week

kimpers:
I tried applying this patch and I am not getting the full intended behaviour.

Scenarios
* App is in the foreground: `onNotification` is called correctly (behaviour unchanged)
* App is killed: `onNotification` is called correctly (fixed by this patch)
* App is in background: `onNotification` is never called (unintended behaviour)

Any ideas?

Gp2mv3:
We merged PR #807 which encloses this one so I close this PR.

PR #806, opened 2018-07-31
Fix title and message for Firebase

steelbrain:
`title` and `body` are special props when coming from GCM/Firebase compat. Their keys are prefixed with `gcm.` which is ignored when `getData()` returns it's data.

We have to use the remote notification APIs to get access to those properties.

I've added these new props before iterating the data bundle so data can override this, making this change backward-compatible.


schumannd:
can the usage of title and body in remote notifications be explained in the README? I couldn't find a reference to it.

steelbrain:
@schumannd The title and body have to be specified in the Firebase payloads

schumannd:
so it does not work with GCM? What about the 'message' key?

steelbrain:
Not sure about GCM but this PR is specific to FCM

PR #807, opened 2018-07-31
Fix pop initial for firebase

steelbrain:
A cleaned up version of https://github.com/zo0r/react-native-push-notification/pull/793

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotification.java
@@ -61,9 +61,18 @@ public class RNPushNotification extends ReactContextBaseJavaModule implements Ac
         return constants;
     }
 
-    public void onNewIntent(Intent intent) {
+    private Bundle getBundleFromIntent(Intent intent) {
+        Bundle bundle = null;
         if (intent.hasExtra("notification")) {
-            Bundle bundle = intent.getBundleExtra("notification");
+            bundle = intent.getBundleExtra("notification");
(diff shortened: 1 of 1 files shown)


Gp2mv3:
Hey, did you check it with the example app in the `example` folder of the repo ?

steelbrain:
@Gp2mv3 For some reason the `example` app doesn't run for me, not even in `master` so if someone else could please try, that would be great.



@akbokha Thank for the other PR btw! It was the base for building this one

steelbrain:
Apologies, I had another bundler running. Good catch!

The test app works, I've tested the new changes with in different scenarios. Here are the results from my testing:

- [x] When a previously closed app is launched, a new JS context is created and initial pop works
- [x] When a notification is clicked from when the app was minimized, JS execution is resumed (JS is not restarted) so it just triggers `onNotification` like expected
- [x] When the app is running and in foreground, `onNotification` is triggered as expected

I did notice that sometimes people override `onNewIntent` in their `MainApplication` in their RN projects, and in that new overriden function they do something like `setIntent(newIntent)`, which does NOT trigger RN new intent handlers and therefore breaks this module.

To the people I've who do what I've mentioned above, they should instead use `super.onNewIntent(newIntent)` instead and then write their custom logic below that so RN can properly fire callbacks. I think this is what is going on in https://github.com/zo0r/react-native-push-notification/pull/793#issuecomment-409276507

Gp2mv3:
Don't worry, I already did the mistake a lot of time. :wink:

If everything goes well, we can merge !

akbokha:
Nice work guys! :)

Issue #814, opened 2018-08-04
Add ability to configure channel name and description


steelbrain:
Currently it's hard-coded as `rn-push-notification-channel` and there's no way to configure it :(

See https://github.com/zo0r/react-native-push-notification/blob/0bcc9ac229528626a2ab94efa0dec0f17358393e/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L487

steelbrain:
Fixed by https://github.com/zo0r/react-native-push-notification/pull/822

tuononh:
It's not work and I have to manually modify `RNPushNotificationHelper.java`.

steelbrain:
Hi @tuononh 

This change has not yet been released. Can you please share how you're using it? Also can you also share what you tried and is not working?

tuononh:
@steelbrain: I hard-coded it in `RNPushNotificationHelper.java`. I tried to modify manifest file as the doc says but it didn't work.

steelbrain:
@tuononh README only works for the latest unreleased stuff, you will have to wait until this is released https://github.com/zo0r/react-native-push-notification/blob/aaf2d19925db4666d6018c0ab1ff0d7d4c90a21d/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationConfig.java#L36

huangkaiw3n:
@steelbrain I'm running into this issue as well. Any estimate for the next release? I'm considering my options before making a production update.

PR #815, opened 2018-08-04
Add channel configuration for Android

steelbrain:
Fixes https://github.com/zo0r/react-native-push-notification/issues/814

Adds ability to set name and description for new Android Oreo notification channels


steelbrain:
I wonder if we can use the Application metadata to specify this, kinda like how firebase does it's config for app 🤔

steelbrain:
Closing in favor of the cleaner approach in https://github.com/zo0r/react-native-push-notification/pull/822

PR #826, opened 2018-08-13
updated sdk and build tool version on gradle file

receme:
This will solve error  “Cannot find symbol on import android.app.NotificationChannel” line that is occurring at runtime.

NotificationChannel is introduced later than 23. So it should be fixed.


joeabill:
💯 I need this so I can upgrade to the latest of RN. Someone please merge! 🙏

Issue #828, opened 2018-08-13
Enable firebase-first messaging


rhdeck:
GCM is on the way out, and mixing GCM with FCM causes issues (the zzi to zzf problem). 

Let's update the package to be FCM-first by making it use the FCM code for getting tokens and receiving. This means introducing support for the google-services.json file as well, which should be importable via command line (using an rnpm plugin seems like the way to do it.)

Issue #827, opened 2018-08-13
Allow automatic deployment 


rhdeck:
Right now the library requires manual edits to the Android Manifest and a couple build.gradle files. Let's add some rnpm link directives to allo deployments to happen automatically.

Gp2mv3:
I just saw your issue and PR. Why did you revoke it finally ? Wasn't it working "as-is" ?

rhdeck:
I've made a new PR - #836. There was a dependency on `react-native-update-gradle` that I had not included in the prior package.

PR #829, opened 2018-08-13
Making slimmer, firebase-first and automatic deployment

rhdeck:
Solving issue #827 and issue #828 by:

1. Removing extraneous iml files (these are build files, not necessary for deployment)
2. Adding the manifest directives to the module's AndroidManifest.xml, which gets merged into the app anyway (removing manual step)
3. Making the code firebase-first, using the modern Firebase techniques to manage tokens (both initial registration and receiving new token updates) 
4. Allow deployment of google-services.json via react-native plugin:
```
react-native add-google-services /path/to/downloaded/google-services.json
```
5. (Bonus) Adding ability to set the application ID of the app in the bundle, which is what matters for the purpose of synch with FireBase. Does not require any changes to Java/Kotlin or JS code. 
```
react-native set-application-id com.me.myfcmtest
```


TikiTDO:
I had to roll back the `AndroidManifest.xml` and `build.gradle` changes in order to get it to compile. I think those make too many assumptions about the environment they're running in.

I'm also not sure if the deployment tools are really a good fit for this lib. That feels like something each project should handle on its own.

rhdeck:
Thanks! What was the environment you were trying to build in?

TikiTDO:
It's a RN app that's been in the works for a couple of years. At this point it's a rather unstable mountain of code. Basically, it's the worst case scenario as far as a lib like this is concerned; apps like the one I'm working on are probably one of the main reasons the config instructions involve manual steps.

TikiTDO:
I don't actually  have any relation with this lib. I just used this issue as a jumping off point to help debug a different bug, and wanted to offer some feedback. Based on the commit history of this repo you're looking to talk to @Gp2mv3 here

rhdeck:
Revoking this PR. Its unwieldly and is still missing a key line in package.json (it has a peer dependency on `react-native-update-gradle` that is not addressed)

PR #848, opened 2018-08-27
Fix app name to be localized in notifications

neospirit:
The title of the content with the app name is not localized.
The key `app_name` allows the title to be localized.


Gp2mv3:
This isn't good as you don't always want to display the app name as title of your notifications.

PR #849, opened 2018-08-27
Android API 26+ fixed heads-up notifications

btav:
Starting in Android API 26+ the notification channel needs **high importance** in order to display heads-up notifications. We upgraded our app from target API 22 to 27 and noticed that the heads-up notifications no longer worked for our local notifications. In this pull-request, I just changed the **NotificationChannel** importance to **high** instead of **default**. This resolved the issue. 

Please review the **heads-up notification** section. 
https://developer.android.com/guide/topics/ui/notifiers/notifications

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -492,7 +492,7 @@ public class RNPushNotificationHelper {
         if (manager == null)
             return;
 
-        int importance = NotificationManager.IMPORTANCE_DEFAULT;
+        int importance = NotificationManager.IMPORTANCE_HIGH;
         NotificationChannel channel = new NotificationChannel(NOTIFICATION_CHANNEL_ID, this.config.getChannelName(), importance);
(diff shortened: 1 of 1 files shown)


sennevc:
From the documentation you linked:

> - Urgent: Makes a sound and appears as a heads-up notification.
> - High: Makes a sound.
> - Medium: No sound.
> - Low: No sound and does not appear in the status bar.

So I think you need to set it to **urgent** instead of **high**.

vmary2014:
I would have thought this need parameterized but for my own use this change would be perfect ;o)

btav:
@sennevc I have only tested using the `high` channel importance and it worked. Sorry for any confusion but I meant this part of the documentation. 

screenshot at aug 28 10-45-00

@vmary2014 Good idea, adding a parameter for the **channel importance** might be a better solution. I can most certainly look into doing this change and updating the documentation. Im just going to wait for one of the maintainers/collaborators to review and give us their thoughts.

sennevc:
Whoops, my bad! I confused two parts of the documentation. ¯\\\_(ツ)_/¯

Gp2mv3:
Shouldn't the importance be a parameter per notification ? (In the payload)

PR #854, opened 2018-08-29
[Android] Allow to set notification's priority, visibility and importance options

lorenc-tomasz:
[Android] Allow to set notification's priority, visibility and importance options
[Android] Change default priority to high
[Android] Change default importance to high


lorenc-tomasz:
With the latest changes we use it on production, however I would be grateful for any additional tests.

PR #859, opened 2018-09-04
Fix README and foreground notifications

prih:
if the application is in foreground = true, then you do not need to send notification to NotificationCenter, only callback in JS. If this is necessary for the application, it can call localNotification

WUT?


mmeshkatian:
its nice really
but better to define a setting for this

prih:
@mmeshkatian I think that this is not necessary because if the application received notifications, then it can, if necessary, send it to the notification center, with this API gives more opportunities for customization for local ones

Dallas62:
Hi @prih 
It's late, but is it the same behaviour as iOS ?
Thanks!

Dallas62:
Hi, every changes are now applied.

PR #878, opened 2018-09-21
Fix: could not find main:"index" in package.json

richie-south:


Dallas62:
Thanks for this PR, nice catch !

PR #881, opened 2018-09-25
Navigate to App notification settings for Android

Taylor123:
## Reasoning 

Within `iOS` there is currently a way to make a call to action for the user to change their push notification permissions for the application in use (happens with the initial `requestPermissions` call). In order to help `android` users proactively change their push notification settings I added a `navigateToAndroidSettings` function.

## Expected Behavior

`iOS`: calling `navigateToAndroidSettings` should be a noop
`android`: calling `navigateToAndroidSettings` will open the settings application to the notification settings of the application it was called from

## Testing

Feel free to call the function directly in `android` or `iOS` and check that the expected behavior above is achieved. 

P.S. apologies for the older commit and merge, meant to make this PR a few weeks ago :)


Dallas62:
Hi @Taylor123 
Long time after you sent the PR, but I will look to implement this ASAP !
Thanks !

PR #889, opened 2018-10-04
Allow monthly repeat

maxkomarychev:
Mimic iOS behavior when last day of month is treated properly


maxkomarychev:
@zo0r @Gp2mv3 bump :)

PR #916, opened 2018-10-24
Fix importance level always set to high

jeepee12:
When creating the NotificationChannel, it was reading the important level from a new Bundle instead of using the one that already exist.
I just added to the method checkOrCreateChannel, the bundle in parameter since its only used from one place and from there you already have the bundle.


jeepee12:
@lorenc-tomasz Can you review this since you created this feature. Thank you

lorenc-tomasz:
It looks ok to me.

tahins:
Why isn't this PR merged already?

Dallas62:
The fix has been applied, thanks for this PR.

Issue #884, opened 2018-09-28
Android >= 8.0 is not receiving notifications


rafaelcamaram:
Yesterday I notice that devices that are using an Android version greater than 8.0 were not receiving push notifications. When I was debugging it, I notice that it's caused by the null value returned by `this.config.getChannelName()` on the `RNPushNotificationHelper.java` file.

A workaround that I found was to make a ternary there in order to set a default value to it if it returns null.

`NotificationChannel channel = new NotificationChannel(NOTIFICATION_CHANNEL_ID, this.config.getChannelName() != null ? this.config.getChannelName() : "rn-push-notification-channel", importance);`

It's working but I know that it's not the best solution for it. Can we improve it? Is there any drawback related to this workaround?

The fork that I'm using with the workaround: https://github.com/rafaelcamaram/react-native-push-notification

Cheers

moneynguyen0o0:
+1

rafaelcamaram:
One thing that I forgot to mention is that this `ChannelName` is required starting in Android 8.0 (API level 26); so all notification must be assigned to a channel. 

But there is no specific name we must put there, you can give the name you want to your channel; it's used only to group all the notifications of the same app - or type - in order to set some properties on it. So, since we're not setting properties on it, I believe that the workaround that I suggested above won't cause any problems to your app until the repository maintainers fix this problem

rafaelcamaram:
One thing that I forgot to mention is that this `ChannelName` is required starting in Android 8.0 (API level 26); so all notification must be assigned to a channel. 

But there is no specific name we must put there, you can give the name you want to your channel; it's used only to group all the notifications of the same app - or type - in order to set some properties on it. So, since we're not setting properties on it, I believe that the workaround that I suggested above won't cause any problems to your app until the repository maintainers fix this problem

rafaelcamaram:
One thing that I forgot to mention is that this ChannelName is required starting in Android 8.0 (API level 26); so all notification must be assigned to a channel.

But there is no specific name we must put there, you can give the name you want to your channel; it's used only to group all the notifications of the same app - or type - in order to set some properties on it. So, since we're not setting properties on it, I believe that the workaround that I suggested above won't cause any problems to your app until the repository maintainers fix this problem

JonesN3:
Did you make a PR for this @rafaelcamaram ?

sonicsmith:
Just tried this now and it works for me! Thanks @rafaelcamaram!

[...]

PR #939, opened 2018-11-15
Android >= 8.0 is not receiving notifications

rafaelcamaram:
I created an issue a few weeks ago because I was not getting notifications on my Android devices. I realized that this happened because the X method was not returning any value to the channel name.

This `ChannelName` is required starting in Android 8.0 (API level 26); so all notification must be assigned to a channel. it's used only to group all the notifications of the same app - or type - in order to set some properties on it.

So I made a simple ternary to make sure there was a value. I know that it's just a workaround. It would be more interest if we could fix it inside the `this.config.getChannelName()`.

I believe that, even though this is only a workaround, I think it would be important to approve this PR (or find some permanent solution) as several people are experiencing this problem - as you can see on the link below.

Issue: [LINK HERE](https://github.com/zo0r/react-native-push-notification/issues/884)


arekzaluski:
Device: Pixel 2
Android version: 8.1.0
React Native version: 0.57.1

Scheduled local notifications are appearing again with this changes. I'm however not receiving `notificationActionReceived` event after clicking an action button. @rafaelcamaram does it work for you?

legrandelf:
@zo0r would you please find time and review this PR, please. It's a really small fix, but a significant one.

alexdor:
@Gp2mv3 @zo0r any updates on this one? It's a really important bug

PR #954, opened 2018-11-30
fix fcm isForeground

zhangjunhou:


zhangjunhou:
Firebase send push notification twice

prih:
mb #859

PR #959, opened 2018-12-07
[Android] Allow to set the notification to executes on idle

hshiraiwa:
On Android 6.0 (API level 23) and above in order to reduce battery consumption, the system enters in Doze mode while the device is unused for long periods of time.
While on Doze mode the AlarmManager alarms dont execute exactly on the scheduled time, they are deferred to the next maintenance window. This cause the notifications to show after their scheduled time.
This change adds a new attribute `allowWhileIdle` to allow the notification to be displayed while on doze

This behavior is documented on: https://developer.android.com/training/monitoring-device-state/doze-standby


cabelitos:
👍

PR #960, opened 2018-12-08
compile statement changed to implementation in build.gradle

akshayjai1:


Dallas62:
Thanks for this PR!
Change is now on master

PR #963, opened 2018-12-14
Updated gradle configuration

hussainahmad:
use `implementation`  instead of `compile`


Dallas62:
This changes are made on master.

PR #1021, opened 2019-03-18
bug fix related to NotificationChannel in android 8.1, API 26

medAmineEld:
Fixing issue of handling notification in android 8.1
the library works great for older version < API 26, but when it cams to API 26 or higher ,notifications does not display anymore.

So i solved this problem after spending several days looking for a solution , i changed some line of code in RNPushNotificationHelper.java file, such as testing if NotificationChannel channel is already created or not .
I posted the solution in SOF [here](https://stackoverflow.com/questions/55214424/react-native-push-notification-doesnt-work-in-android-8-1-api-level-27)


Dallas62:
Hi!
Thanks for this PR, this change is on the code base 😉

Issue #843, opened 2018-08-24
JSON value '<null>' of type NSNull cannot be converted to NSDictionary


vikrantnegi:
# Description
I'm using the example project files in my project to test the notifications and it is working fine in Android.

But on iOS, I'm getting this error

`JSON value '<null>' of type NSNull cannot be converted to NSDictionary`

whenever I'm clicking on local as well as schedule notification. Although It is working fine and I'm receiving the notification but I'm getting this error every time I'm clicking on them.

![img_0196](https://user-images.githubusercontent.com/3294287/44571125-8356ad80-a79d-11e8-9062-c222575243ab.jpg)

# Details
"react-native-push-notification": "^3.1.1",
"react": "16.4.1",
"react-native": "0.56.0",

vygandas:
Same here.

efstathiosntonas:
same here

jriiringan:
i tried to fix this by passing an empty object on userInfo instead of null value for notification.

vikrantnegi:
@jriiringan I'm getting the same error. Were you able to fix it using that?

jriiringan:
@vikrantnegi  yes i did. the default example send "null" value so i replace it by an Object.

SCJMENGMENG:
Modify the NotifService's "category" to be a string,'userInfo'to be a dictionary, and'number' to be a number.

anhvd:
@SCJMENGMENG s
thank for your answer, that's true.

hungdev:
@SCJMENGMENG it's worked for me. thank u.

e1016:
In my case, for future references, the problemas was a `null` value for `Image uri` getted from server

`<Image source={{ uri: data }}>` <- `data is null`

Dallas62:
I fixed the README and example application.

kovkev:
Has anyone gotten react-native-push-notification working with node-pushnotifications ?

I try to get node-pushnotification to send a push notification to my app, which shows a push notification on ios, but then when I click on it (react-native-push-notification is handling here), then I get the same error as above:

`JSON value '<null>' of type NSNull cannot be converted to NSString`

kovkev:
It was happening because I did 

```
PushNotification.cancelLocalNotification({ id: notification.data.notificationId });
```

PR #1030, opened 2019-03-30
Update NotifService.js

barclayd:
Fixes an error thrown when running the Example using iOS simulator as a notification is sent:
```JSON value '<null>' of type NSNull cannot be converted to NSDictionary```

Fixes the issue raised here: https://github.com/zo0r/react-native-push-notification/issues/843

image


giacomocerquone:
@zo0r please check this, I've lost 20 mins for this thing here

@barclayd please can you modify readme too?

PR #1058, opened 2019-04-26
Fix double notification when app is in foreground

EvanSays:


quinnturner:
Just an FYI, you added too many files to your PR. Only `RNPushNotificationListenerService.java` should be added.

Issue #952, opened 2018-11-28
How to use this library when you have multiple push providers?


shercoder:
I'm using react-native-push-notification along with Localytics push messaging. So I created a custom FirebaseMessagingService that checks if RemoteMessage is a Localytics message else it needs to fire `RNPushListenerService`. 

I'm starting RNPushListenerService using startService but it does not look like it is starting it. Has anyone else dealt with issue like this before? and How did you handle it?

Note: There can only be one FirebaseMessagingService with intent-filter of `com.google.firebase.MESSAGING_EVENT` and I'm registering my custom service with that intent-filter.

```
public class CustomFirebaseMessagingService extends FirebaseMessagingService {
    @Override
    public void onMessageReceived(RemoteMessage remoteMessage) {
        Map<String, String> data = remoteMessage.getData();
        try {
            if (!Localytics.handleFirebaseMessage(data)) {
                forwardPushNotification(remoteMessage);
            }
        } catch (Exception e) {
            Timber.e(e, "Failed to extract Push Message", remoteMessage.getMessageId());
        }
    }

    @Override
    public void onNewToken(String token) {
        super.onNewToken(token);
        Localytics.setPushRegistrationId(token);
    }

    private void forwardPushNotification(RemoteMessage remoteMessage) {
        Map<String, String> data = remoteMessage.getData();
        data.put("from", remoteMessage.getFrom());
        Intent intent = new Intent(this, RNPushNotificationListenerService.class);
        intent.setAction("com.google.firebase.MESSAGING_EVENT");
        Bundle extraData = new Bundle(data.size());
        for (Map.Entry<String, String> entry : data.entrySet()) {
            extraData.putString(entry.getKey(), entry.getValue());
        }
        intent.putExtras(extraData);
        startService(intent);
    }
}
```

I think react-native-push-notification library needs to provide a custom method like the one Localytics is providing to handle these notifications when the user has multiple push providers. e.g. `Localytics.handleFirebaseMessage`

gorjuspixels:
@shercoder 	have you figured this out?

shercoder:
@gorjuspixels Yeah I ended up forking this repo and fixing it in a feature branch. I haven't had a chance to make a PR for this repo. I'll see if I can do that this weekend but here is the draft PR if you need example https://github.com/buildcom/react-native-push-notification/pull/1

shercoder:
Then In your custom FirebaseMessagingService, you can do something like this:
```Java
[...]

PR #1128, opened 2019-07-09
Enables the ability to support multiple push providers

shercoder:
**This change only impacts Android**

The need for this change came because the project I work has a need to support multiple push providers. We have an internal provider where we manager our in-house push notifications and then there is a need for external/marketing push notifications that we depend on a third-party to deliver.

Since there can only be one FirebaseMessagingService with intent-filter of com.google.firebase.MESSAGING_EVENT, we cannot have a service from Localytics and RNPushNotificationListenerService declared in our AndroidManifest. Android by default will only pick the first declaration.

In order to support multiple providers, a developer can create a custom FirebaseMessagingService that distributes the push notification to providers. Please see this issue I created that explains the problem and the solution in more details https://github.com/zo0r/react-native-push-notification/issues/952

Basically this change now allows the developer in need to do the following in their custom FirebaseMessagingService:
```Java
private RNReceivedMessageHandler receivedMessageHandler = new RNReceivedMessageHandler(this);

...
...
...

@Override
    public void onMessageReceived(RemoteMessage remoteMessage) {
        Map<String, String> data = remoteMessage.getData();

        String from = remoteMessage.getFrom();
       
            try {
                if (from first provider) {
                    receivedMessageHandler.handleReceivedMessage(remoteMessage);
                } else {
                    // pass remote message to another provider
                }
            } catch (Exception e) {
                Log.e(e, "Failed to extract Push Message", remoteMessage.getMessageId());
            }
        
    }
```

**NOTE:** There will be no change in behavior in how the current version of this library works and the users who don't need the support for multiple push providers won't have to change anything in their code base and continue to use the `RNPushNotificationListenerService` as it is.

**Testing**
1. For single provider, you can just go through the normal setup as described.
2. For multiple providers, go through the normal setup for this library but instead use your customer `FirebaseMessagingService` instead of `RNPushNotificationListenerService` as mentioned above.


plus-:
This would also let us localise the messages from the push notification (if anyone know a work around for this, please let me know)

nchase:
@zo0r have you seen this PR? Supporting multiple push providers from the npm package would be great 🙂

mysport12:
@Dallas62 it seems like you have been driving the repo forward lately so just wanted to be sure you saw this PR. This functionality would be amazing to have for a number of different projects.

Dallas62:
I will take a look ASAP 😉

mysport12:
> I will take a look ASAP 😉

Awesome! Much appreciated, and also a big thank you to @shercoder for the PR!

shercoder:
Thank you @mysport12 and @Dallas62. This would be a great addition to the lib, one of my former colleague ran into the same problem and I had to point him to my forked branch for the time being.

Dallas62:
Hi @shercoder 
Did you have a bit of time to make your changes compatible with the latest version ?
In this PR or another, if not, I will try to take time to reimplement it on dev branch ASAP.

shercoder:
@Dallas62 I'm not 100% but I'll see if I can carve some time this weekend. Also, I haven't looked at changes in 0.62 (latest), are there any changes I should be aware of?

sdenaci:
Hi! I needed to use this library with another push provider on a project I am currently working on, so took a stab at reimplementing @shercoder 's pr back on current dev. Am submitting it back to library in case it is helpful! Here is the pr https://github.com/zo0r/react-native-push-notification/pull/1445

Dallas62:
Hi,
Since @sdenaci made another PR, I close this one.
Changes will be released soon.

PR #1132, opened 2019-07-09
Update react-native.config.js

chrisspankroy:
Add proper dependency tag to react-native.config.js. Looking at how other repositories handle the update from `rnpm` to `react-native.config.js`, this is the right way to do it.

### DIFF
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -1,7 +1,9 @@
 module.exports = {
-    project: {
-      android: {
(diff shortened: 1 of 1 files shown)


kbrandwijk:
Would be great if this could be merged. It's the only warning in my project after the RN 0.60 migration

theweavrs:
@zo0r ping

PR #1185, opened 2019-08-29
Update README.md

AssisrMatheus:
The readme file has a "typo" on its local notification properties 😅


lmrarturo:
Sorry for the last comment.

Just to say that action is only supported in Android, for iOS actions the docs [recommend this module](https://github.com/holmesal/react-native-ios-notification-actions). (it's an old module but I don't know if it is working well so far).

Dallas62:
Since the conflict is based on indentation, I will apply this directly on `dev`.

PR #1193, opened 2019-09-09
Removing types from vanilla javascript file to fix errors in react native web.

scottsdevelopment:
These typed out functions are breaking when including the modules from my build.

### DIFF
--- a/component/index.android.js
+++ b/component/index.android.js
@@ -26,19 +26,19 @@ NotificationsComponent.prototype.getInitialNotification = function () {
         });
 };
 
-NotificationsComponent.prototype.requestPermissions = function(senderID: string) {
+NotificationsComponent.prototype.requestPermissions = function(senderID) {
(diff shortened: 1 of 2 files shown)


DanielDavidDev:
what is the purpose of this issue, removing the reference of types?

scottsdevelopment:
So the .js file isn't a .ts file, I guess the fix would be to either include documentation about requiring a module loader for the specific path and having it be transpiled at compile time or removing the invalid TS type references in the JS so it just works as-is.

scottsdevelopment:
Something like this in your webpack might also fix the problem:
``` 
      {
        test: /\.js$/,
        use: 'ts-loader',
        include: [
            path.resolve(__dirname, 'node_modules/react-native-push-notification')
        ],
      }
```

Dallas62:
Hi,
This should be ok now.

PR #1042, opened 2019-04-09
[Android] Fixed foreground FCM banner notifications and notification sound

tankers746:

PR #1243, opened 2019-11-14
Custom sound solved in android oreo and higher

cookienawer:
Fixed the problem of custom sounds. With the official version, it works perfectly with versions before oreo, but in the oreo version no audio sounded, and in higher versions the terminal's default audio was sounded.

Now in all versions it works correctly.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -16,6 +16,7 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.media.RingtoneManager;
+import android.media.AudioAttributes;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -598,6 +601,16 @@ private void checkOrCreateChannel(NotificationManager manager) {
...
         channel.enableVibration(true);
 
+        if (soundUri != null) {
+            AudioAttributes audioAttributes = new AudioAttributes.Builder()
+                    .setContentType(AudioAttributes.CONTENT_TYPE_SONIFICATION)
+                    .setUsage(AudioAttributes.USAGE_ALARM)
+                    .build();
+            channel.setSound(soundUri, audioAttributes);
+        } else {
+            channel.setSound(null, null);
```
Dallas62, line 611: What is the purpose of this line ?
Reset the sound if there is not passed in parameter ?

Dallas62:
I just find this PR #1042 
Did you test it ?

PR #1355, opened 2020-04-02
Use FirebaseInstanceId for deviceToken, not from Intent

ls-jack-wong:
Intent does not generate the correct FCM token for release builds. Use the FirebaseInstanceID to get the correct one instead.


Dallas62:
Hi @jackw-kounta 
Thanks for your PR, nice investigation.
I wait some feedback from other issues.

Dallas62 (approved):

ToTheSol:
I'm actually having this issue, and I pulled from the latest version of this package, but I don't see some of the code that's here in the file specified when I look at this in node modules.

lucaszdev:
> I'm actually having the issue where I get remote notifications in debug and not in release, and I pulled from the latest version of this package, but I don't see some of the code that's here in the file specified when I look at this in node modules.

Same Here

PR #1363, opened 2020-04-12
Missing step for android manual installation 

paddy57:


Dallas62:
Hi @paddy57 
Thanks for this PR
Are you sure this step is mandatory ?
As far as I know, the autolink should work on Android
Regards

paddy57:
Hello, @Dallas62 I tried manual Android manual Installation, and after building an android project I got an error saying ReactNativePushNotificationPackag does not exist. And in order to solve, I had to add react-native-push-notification dependency in android build Gradle. So I think it's needed as auto-link does not work for me.

PR #1372, opened 2020-04-16
Add abandonPermission for android

CoverID:


Dallas62:
Hi @CoverID 
Thanks for this commit, is your changes for GCM or FCM ?
I think we should remove GCM in a future release since it's deprecated

PR #1425, opened 2020-05-08
Added Abandon Permissions method to Android

### LINKED ISSUE #864: Reset token
Is it possible to force reseting the token on every logout so that when a new user logs in to the app she receives a new token?

GabeConsalter:
The Abandon Permissions method makes possible to revoke the current token, generating a new one on registration.

Unregister for all remote notifications received via APNS or FCM.

In case many users could be using the same device, Revoke Token will be useful to prevent notifications to the previous user from the device, using it in a logout action for example.

Fixes #864

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,6 +413,11 @@ Removes the specified notifications from Notification Center
 | ----------- | ----- | -------- | ---------------------------------- |
 | identifiers | array | Yes      | Array of notification identifiers. |
 
+
+## Abandon Permissions
+
+`PushNotification.abandonPermissions()` Revokes the current token and unregister for all remote notifications received via APNS or FCM.
+
 ## Notification priority
 
 (optional) Specify `priority` to set priority of notification. Default value: "high"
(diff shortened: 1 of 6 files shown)


Dallas62:
Hi @GabeConsalter 
Thanks for this PR, maybe we can rename the method to:
```js
PushNotification.abandonPermissions()
```
The API will be the same for Android and iOS.

Dallas62:
I will test it ASAP.

Dallas62:
Released in `3.4.0` due to another fix, no need to reload the application to get the new token 😉

GabeConsalter:
All right, thank you @Dallas62

Pablo14697:
Hi everyone! I think that there is some weird like a default behavior on Android platform. I'm talking about when I create a new RN project and after that I install FCM and react-native-push-notification with the common comands:
`yarn add @react-native-firebase/messaging react-native-push-notification`

WITHOUT one step more I CAN START TO RECEIVE NOTIFICATIONS. So, I'm new in this library but that is normal? If that's right how can i disable that default behavior and only use the @react-native-firebase/messaging listeners and send a push notification when the listeners get some notification?

PR #1441, opened 2020-05-15
Add missing push timestamp for Android 7+

xvonabur:
I noticed that push notifications received on Android do not show sent date timestamps. 
According to [official docs](https://developer.android.com/reference/android/app/Notification.Builder.html#setShowWhen\(boolean\)) it's off for Android 7 and higher.

Here is example:

This PR enables it by default for every Android version.


Dallas62:
I will allow developper to override this in options 😉

PR #1466, opened 2020-06-05
Added method getScheduledLocalNotifications()

lukebars:
This method is really useful for organizing notifications, e.g. cancelling some of already scheduled ones. Solves #673

For iOS it uses [already existing method](https://github.com/react-native-community/push-notification-ios#getscheduledlocalnotifications)

I've written a method for android which does the same.

PR #1473, opened 2020-06-09
Updating push-notification-ios to version 1.2.2 to fix some issues with the installation itself.

willnaoosmit:


Dallas62:
Hi @willnaoosmit
Thanks for this PR, since this change will be made in https://github.com/zo0r/react-native-push-notification/pull/1466, I will close it and use PR 1466 in next release.

PR #1474, opened 2020-06-10
Add notification id to iOS if userInfo doesn't exist

lukebars:
Currently we add notification `id` variable to notification object on iOS only if `userInfo` exists. 

If `userInfo` is not set, `id` is not being set. 

This PR fixes that.

Cheers 🍻


lukebars:
cc @Dallas62

Dallas62:
Nice !
Thanks !

PR #1475, opened 2020-06-10
Return a Promise on getScheduledLocalNotifications

lukebars:
I've noticed that current concept of passing callback to the `getScheduledLocalNotifications` function as an argument wasn't very smart or user friendly, and a nightmare in redux sagas.

Returning a promise keeps it way more simple.

Correct me if I'm wrong @Dallas62

### DIFF
--- a/example/App.js
+++ b/example/App.js
@@ -108,7 +108,7 @@ export default class App extends Component {
         <TouchableOpacity
           style={styles.button}
           onPress={() => {
-            this.notif.getScheduledLocalNotifications(notifs => console.log(notifs));
+            this.notif.getScheduledLocalNotifications().then(notifs => console.log(notifs));
           }}>
           <Text>Console.Log Scheduled Local Notifications</Text>
         </TouchableOpacity>
--- a/index.js
+++ b/index.js
@@ -452,40 +452,46 @@ Notifications.getDeliveredNotifications = function() {
   return this.callNative('getDeliveredNotifications', arguments);
 }
 
-Notifications.getScheduledLocalNotifications = function(callback) {
-	const mapNotifications = (notifications) => {
-		let mappedNotifications = [];
-		if(notifications?.length > 0) {
-			if(Platform.OS === 'ios'){
-				mappedNotifications = notifications.map(notif => {
-					return ({
-						soundName: notif.soundName,
-						repeatInterval: notif.repeatInterval,
-						id: notif.userInfo?.id,
-						date: new Date(notif.fireDate),
-						number: notif?.applicationIconBadgeNumber,
-						message: notif?.alertBody,
-						title: notif?.alertTitle,
-					})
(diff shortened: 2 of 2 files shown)


lukebars:
@Dallas62 I think moving towards promises would be a lot better solution. Currently it's kind of asynchronic but you can't use it in a proper way.

Imagine having loads of scheduled notifications that must be managed on further user interaction. E.g. you want to cancel some of the notifications conditionally by their id:

```
 const notifs = yield call(Notifications.getScheduledLocalNotifications);
    const regex = action.payload;
    const regexp = new RegExp(`^(${regex})`);

    notifs.forEach(notif => {
      if (String(notif.id).match(regexp)) {
        Notifications.cancelNotif(notif.id);
      }
    });
```

Now imagine you have some more logic. This wouldn't make any sense to pass the whole logic inside the callbacks IMHO.

Dallas62:
Of course promises can simplify the way you handle your logic.
But you probably need to apply this change to all methods with callback and not only one.

What a tried to point out is we can wrap the method with callback in a promise, if someone really need this.

Maybe we can start by listing all methods with callback.

lukebars:
@Dallas62 good insights. I'll take a look into that.

lukebars:
Currently there is only 4 methods that use callback:
`checkPermissions`
`popInitialNotification`
`getApplicationIconBadgeNumber`
`getScheduledLocalNotifications`

I think there are two solutions:
1. Wrap a method in a promise with a callback inside.
2. Provide different named methods for promises and callbacks.
e.g.

```
export const getSomethingAsync = () => new Promise(...)

export const getSomething = (callback) => {
    getSomethingAsync.then(callback);
}
```

Dallas62:
- [ ] checkPermissions
- [ ] popInitialNotification
- [ ] getApplicationIconBadgeNumber
- [ ] getScheduledLocalNotifications
- [ ] getDeliveredNotifications
- [ ] getChannels

I think we can apply the first solution which is compatible with the "old"-way and avoid this duplicate method with `Async`.
But I'm not sure if this the right way to implement it (double call to resolve/callback).

This can be implemented this way:

```js
function getSomething(callback) {
  if (typeof callback === 'undefined') {
    return new Promise(function (resolve, reject) {
      getSomething(function (err, result) {
        err ? reject(err) : resolve(result)
      })
    })
  }

// ... function logic
}
```

In our case, I'm not sure the first argument of the callback is `err`, so we go for:

```js
function getSomething(callback) {
  if (typeof callback === 'undefined') {
    return new Promise(function (resolve, reject) {
      getSomething(resolve(result));
    })
  }

// ... function logic
}
```

lukebars:
@Dallas62 My second thoughts would be that it's not really needed to add promises to these methods, as users can write promise wrappers in their `NotifService.js`

PR #1509, opened 2020-07-07
Add function createChannel for custom Android channel support

aranda-adapptor:
The rationale behind this PR is that the current functionality with default channel names does not look very professional the end user. Apps can use this new function along with the Manifest setting to disable default channel creation to have fully customised channels.

This has had a fair bit of internal testing and the app using it will be updated for the public soon. I realised after doing this that the `importance` parameter should have used the string mappings already existing, but I don't have time to update that at the moment. Also the typings have not been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,23 @@ In the location notification json specify the full file name:
 
 ## Channel Management (Android)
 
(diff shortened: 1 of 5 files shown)


Dallas62:
Hi @aranda-adapptor 
Thanks for this PR, it's really helpful and was in the backlog ❤️ 
Like others PRs, will test it ASAP, hopefully this weekend

PR #1510, opened 2020-07-07
Add Android messageId to enable integration with react-native-firebase/messaging

aranda-adapptor:
This PR is part of the solution to the following problem:
- Using `@react-native-firebase/messaging` for delivery and handling or remote messages and notifications when the app is killed or in the background
- Using `react-native-push-notification` to display these notifications when the app is in the foreground
- Tapping on a push that arrived in the foreground opens the app but does _not_ send the payload through to RN firebase `messaging().getInitialNotification` or `messaging().onNotificationOpenedApp`

RN firebase messaging stores the remote notification payload and looks it up based on this "message_id" intent extra. The full solution requires minor changes to RN firebase too, which will be submitted as another PR soon.


mikehardy:
The related PR in react-native-firebase has passed the required number of reviews, so it will merge just pending @aranda-adapptor signing the CLA there, so as this is a cross-package PR basically, just keeping you guys up to date that this is good on our end. Cheers

Dallas62:
Hi @aranda-adapptor 
Thanks for this PR !
Changes looks good for local notification, but it' probably not working for Scheduled Notification.
Changes should be made in this file:
https://github.com/zo0r/react-native-push-notification/blob/dev/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java
Once it's done, This PR will be OK for merging.

Dallas62:
Will add the change for Scheduled notifications, thanks !

aranda-adapptor:
@Dallas62 thanks for finishing that off, I was on vacation and it would've taken a little while to figure out the required changes 👍

PR #1511, opened 2020-07-07
Support empty strings as default push notification icons

aranda-adapptor:
The reason for this PR as that when using `@react-native-firebase/messaging`, background notifications show up with different icons to `react-native-push-notication`. For backwards compatibility, I've added support for empty strings causes no icon customisation on the native notification, which mimics behaviour of the RN firebase messaging module.


Dallas62:
Hi @aranda-adapptor 
Will test this ASAP !

PR #1516, opened 2020-07-08
Populate userInfo on android, update RNCPushNotificationIOS

lukebars:
This aligns the object returned by `onNotification`.

Currently `onNotification` returns:
```
{
   "android":{
      "android.intent.extra.ALARM_COUNT":1,
      "autoCancel":false,
      "finish":[],
      "fireDate":1594216704146,
      "foreground":false,
      "id":"2941",
      "largeIcon":"ic_notification",
      "message":"message",
      "notificationId":2941,
      "smallIcon":"ic_notification",
      "title":"hello",
      "userInfo":{
         "screen":"welcome"
      },
      "userInteraction":true
   },
   "ios":{
      "alert":"message",
      "badge":0,
      "data":{
         "id":"6380",
         "screen":"welcome"
      },
      "finish":[
         "Function finish"
      ],
      "fireDate":"2020-07-08T16:59:14.916+03:00",
      "foreground":true,
      "id":"undefined",
      "message":"message",
      "sound":"default",
      "userInteraction":false
   }
}
```

As you can see iOS returns `userInfo` as `data`, and android returns `userInfo` as `userInfo`. Also - id is not being correctly populated on iOS.

My proposal would return this object `onNotification`:
```
{
   "android":{
      "android.intent.extra.ALARM_COUNT":1,
      "autoCancel":false,
      "data":{
         "id":"9683",
         "screen":"welcome"
      },
      "finish":[
         "Function finish"
      ],
      "fireDate":1594219524047,
      "foreground":false,
      "id":"9683",
      "largeIcon":"ic_notification",
      "message":"message",
      "notificationId":9683,
      "smallIcon":"ic_notification",
      "title":"hello",
      "userInfo":{
         "screen":"welcome"
      },
      "userInteraction":true
   },
   "ios":{
      "alert":"message",
      "badge":0,
      "data":{
         "id":"2593",
         "screen":"welcome"
      },
      "finish":[
         "Function finish"
      ],
      "fireDate":"2020-07-08T17:41:44.910+03:00",
      "foreground":true,
      "id":"2593",
      "message":"message",
      "sound":"default",
      "userInteraction":false
   }
}
```

This would also allow us to provide data when getting scheduledNotifications, which I would implement little bit later. P.S. currently on iOS alert(title) is wrong, that's because iOS doesn't return notification title. I will raise a PR in RNC repo for that. Naming should be tweaked too (android - title, iOS - alert). Another possible solution would be naming `data` as `userInfo`


lukebars:
cc @Dallas62

Dallas62:
Hi @lukebars 
Missed your PR in tons of notifications 😅
Thanks for this PR, changes looks great!
I will look into it ASAP 😉

lukebars:
@Dallas62 did you manage to look into it?

Dallas62:
Will check this this afternoon

Dallas62:
Hi @lukebars 
What's strange is that there is no userInfo in the iOS return

lukebars:
Hmmm. I'll check that in few hours. Very strange.

lukebars:
@Dallas62 thing is that iOS returns `userInfo` as `data`, so we should choose do we return it as `data` to keep consistency between this lib and ios one, or should we return it as `userInfo`

Dallas62:
I will take a close look to this, because it's strange to set `userInfo` in iOS, use it every where but return it as `data`.
Busy this weekend, will dig into it on monday.

Dallas62:
> @Dallas62 thing is that iOS returns `userInfo` as `data`, so we should choose do we return it as `data` to keep consistency between this lib and iOS one, or should we return it as `userInfo`. Also we could return both, but that wouldn't make sense..
> 
> I could remove `userInfo` from Android to keep it the same way as in iOS

Hi @lukebars 

I'm not totally convinced by this kind of "renaming" as an optimal solution, but it's probably the best way to unify Android and iOS actually. So, no problem to remove `userInfo` from Android and keep it the same way as in iOS.
We should probably check that Scheduled Notification accept `userInfo` and that there is no mix of naming when the application start in Android.

In the future, I think we should rewrite the API to unify and clean it. Also make a clear documentation. 😄

lukebars:
@Dallas62 or we can use `userInfo` on both platforms instead of `data`. That would be more clear.

Dallas62:
Not sure it's better, because this is the name of the attribute of remote notifications.

lukebars:
Ok, sure. 
I've raised a PR for getting alertTitle in iOS @RNC, as soon as it's merged I'll push some changes here to keep it straightforward.

Dallas62:
Thanks a lot! 🙏

lukebars:
Right. So I've bumped the `@react-native-community/push-notification-ios` to `1.4.0` and aligned some of the stuff returned on `onNotification`. Here's the current object being returned on both platforms:

```
{
   "android":{
      "android.intent.extra.ALARM_COUNT":1,
      "data":{
         "id":1,
         "screen":"home"
      },
      "finish":[
         "Function finish"
      ],
      "fireDate":1595316704887,
      "foreground":false,
      "id":"1",
      "message":"My Notification Message",
      "notificationId":1,
      "number":"10",
      "playSound":false,
      "soundName":"default",
      "title":"Scheduled Notification",
      "userInfo":null,
      "userInteraction":true
   },
   "ios":{
      "badge":10,
      "data":{
         "id":1,
         "screen":"home"
      },
      "finish":[
         "Function finish"
      ],
      "fireDate":1595316681330,
      "foreground":true,
      "id":1,
      "message":"My Notification Message",
      "soundName":null,
      "title":"Scheduled Notification",
      "userInteraction":false
   }
}
```

index.js
```diff
@@ -336,32 +336,28 @@ Notifications._onNotification = function(data, isFromBackground = null) {
...
+        fireDate: Date.parse(data._fireDate),
         finish: (res) => data.finish(res)
       });
     } else {
       var notificationData = {
         foreground: ! isFromBackground,
         finish: () => {},
-        ...data
+        ...data,
+        data: data?.userInfo,
```
Dallas62, line 357: Hi @lukebars 
I'm not sure on this line, I will test on local but this is probably overriding the data field of FCM.

lukebars:
Hey @Dallas62, I've updated the data population, currently it keeps FCM data. Looks good to me. Can you review it one last time? 😄

Dallas62:
Hi @lukebars 
I change the order of data/userInfo for Android and I moved the process in another function. That allow popInitialNotification to follow the same rule.
I need to check for Android now.

lukebars:
Looks good :shipit:

lukebars:
Are we ready for some sort of release?

Dallas62:
Hi @lukebars 
I'm waiting for feedback from @mathiasmoeller on a fix introduced in this PR. (#1501)
I will re-check it this week and merge ASAP.

lukebars:
Hey @Dallas62, I've noticed a strange behavior on android that comes from `handlePopInitialNotification`.

When app is launched from killed state / fresh launch, and you schedule a first notification, `onNotification` fires twice for the first one. Native listener fires only one. 

I see that if you comment out `this._onNotification(firstNotification, true);`, this behavior stops.

Was it intentional to do this or is it a bug?

Dallas62:
@lukebars This is strange, I don't see this behavior

lukebars:
@Dallas62 hmm... seems like this is happening because I store my notification setup not in App.js constructor, but in separate file and import it for my usage like this:

**Usage:**
```
import {notif} from './NotifSetup';

...
notif.scheduleLocalNotification(...);
...
```

**NotifSetup.js:**
```
import NotifService from './NotifService';

const onRegister = token => {
  console.log({ registerToken: token.token, fcmRegistered: true });
};

const onNotif = notification => {
  console.log('NotificationHandler:', notification);
};

export const notif = new NotifService(onRegister, onNotif);
```

Any suggestions/ideas why this could be happening for android only? 😄

Dallas62:
Can you test on the example project (without change expect setting the PR), if his is happening this is probably in the library.


In your code, does `Notifications.scheduleLocalNotification(...);` is trigger by an event such as button pressed ?

lukebars:
Hmm @Dallas62 .. Looks like it was happening even before these changes. I've tested it on example with this kind of setup on master and this branch, and it still does the same.

There is no difference if scheduling a notif is triggered by a button or a render, or just a call on top of `App.js`. Clicking one notif results in two `onNotification` calls 😞 

This though only happens on android and only on first notification on app open from killed/install. (popInitialNotification explains that eh).

Dallas62:
Can you give the steps to reproduce it on the example project ?

lukebars:
Here's the most minimal change requiring reproduction:

**App.js**
```jsx
...
import { notif } from './NotifSetup'; //instead of NotifService
...
  constructor(props) {
    super(props);
    this.state = {};

    this.notif = notif; // instead of new notifService
  }
...
```

**NotifSetup.js**
```js
import NotifService from './NotifService';

const onRegister = token => {
  console.log({ registerToken: token.token, fcmRegistered: true });
};

const onNotif = notification => {
  console.log('NotificationHandler:', notification);
};

export const notif = new NotifService(onRegister, onNotif);

```

Dallas62:
I don't see any problem, does the log:
```js
console.log('NotificationHandler:', notification);
````
inside NotifSetup is right ? it's written `NotificationHandler`

**Comment by lukebars** [comment]
> I don't see any problem, does the log:
> 
> ```js
> console.log('NotificationHandler:', notification);
> ```
> 
> inside NotifSetup is right ? it's written `NotificationHandler`

Didn't it result in double log when clicked on notification?

**Comment by Dallas62** [comment]
Yes it's logged twice the notification, but the example Application has the same log in NotificationHandler and in the NotifSetup you provided, which explain this behavior

**Comment by lukebars** [comment]
@Dallas62 Oh I was wrong about that 🤦 

I've actually managed to reproduce it now. We're using branch in our project for analytics and it requires setting intent.

Inside `example/android/app/src/main/java/com/example/MainActivity.java`:

```java
package com.example;
// Add this:
import android.content.Intent;

public class MainActivity extends ReactActivity {

  // Add this:
  @Override
  public void onNewIntent(Intent intent) {
      super.onNewIntent(intent);
      setIntent(intent); // this is what causes double onNotif call on fresh launch
  }

// Everything else
...
```

lukebars:
@Dallas62 in `index.js:119`:

Changing this
```
  AppState.addEventListener('change', handlePopInitialNotification.bind(this));
```

to this:

```
  AppState.addEventListener('change', (state) => {
    handlePopInitialNotification.bind(state)
  });
```

fixes the double `onNotification` call

Dallas62:
Hi @lukebars 
This change is wrong, the first parameter of `.bind()` is context. By changing this, the `onNotification` un the listener will never trigger.

A correct change would be:
```js
  AppState.addEventListener('change', (state) => {
    handlePopInitialNotification(state);
  });
```
But you will see the same result.

There is a protection to block double trigger based on `notification.id`. Do you see any id in the notification ?

lukebars:
@Dallas62 eh, my bad. 

Both onNotification messages fire with `"id": "1"`.

Looks like  if `setIntent` is configured in `MainActivity.java`, `popInitialNotification` is only needed for iOS, because `onNotification` fires even with listener commented out.

Dallas62:
What is strange is that you are talking about double trigger even when you comment this._onNotification(firstNotification, true);

But it's impossible to get `onNotification` triggered whitout this line. So it can't be trigger twice and only once after removing this line.

lukebars:
Nope. When  that line is commented out, it get triggered only once on Android.

Have you tried adding 

```
  @Override
  public void onNewIntent(Intent intent) {
      super.onNewIntent(intent);
      setIntent(intent);
  }
```

to `MainActivity.java` ?

Instant double call without any code modification inside the lib.

Dallas62:
I'm currently testing it, and I found the reason of that. Making more tests.
popInitialNotification is working fine, this is a problem with onNewIntent inside the library which looks like useless.

lukebars:
That explains everything. Looks like it's doing the same thing as the js initial notif part, only a bit worse. Nice catch 🚀

Dallas62:
Yes but it's probably not so easy. on iOS, clicking on a notification when the Application is opened doesn't trigger the current part of "initialNotification", only Android do that

Dallas62:
Also, the value of `foreground` looks not valid.

Dallas62:
I think it's good now 😉 
Nice investigations !

lukebars:
@Dallas62 awesome work! Tested it out and it looks really good 🚀 

Are we getting close to closing this and releasing it out to the daylight?

Dallas62:
Hi @lukebars 
I was waiting for your feedback, will take a look today 😉

Dallas62:
Done ✅ 

Found the same issue you reported on iOS while testing 😄 

Thanks a lot for this PR !

PR #1518, opened 2020-07-09
Firebase auto init, fix onRegister after abandonPermissions (android)

prih:


Dallas62:
Hi @prih 
I understand the problem here, but this kind of changes in the documentation will lead to copy-and-paste without reading it. Then a tons of issues about "token not generated" will be opened.

PR #1519, opened 2020-07-09
[ANDROID] Add support for setOnlyAlertOnce property

AresMaster:
Add support for setOnlyAlertOnce property. 
From Android documentation:
Set this flag if you would only like the sound, vibrate and ticker to be played if the notification is not already showing.


Dallas62:
Hi @AresMaster 
Thanks for this PR, I'm not sure your change is working with scheduled notification, ca. You check that ?

Dallas62:
I will add the Scheduled part in a future commit, thanks !

PR #1613, opened 2020-08-25
Add help about local notifications problems

Scriptonita:
* describe briefly the rational of the change, why did you do this (to create new feature, bug fix, add some missing docs etc...)?

It adds some help about why users can be facing a strange problem when local notifications only work if user has remote debugger enabled.

* detail the purpose of the change - **what is the expected change in behaviour**?

To help developers that can have this problem.
It is not changing any piece of code

 * state which operating systems are affected

Android and iOS

* provide details on how to test the change

It is not needed

### DIFF
--- a/trouble-shooting.md
+++ b/trouble-shooting.md
@@ -7,6 +7,31 @@ Known bugs and issues:
  * (Android) Tapping an alert in the notification centre will sometimes not result in `onNotification` being called [issue 281](https://github.com/zo0r/react-native-push-notification/issues/281)
  * (Android) Not all local notification features are supported yet (PRs welcome)
  * (iOS) The OS can penalise your app for not calling the completion handler and will stop (or delay) sending notifications to your app. This will be supported from RN-0.38 [PR 227](https://github.com/zo0r/react-native-push-notification/pull/277)
+ * (Android and iOS) Don't use a string to get the date for schedule a local notification, it only works with remote debugger enabled, [explanation](https://stackoverflow.com/a/41881765/8519917).
+  
+
+  ```javascript
+  // It doesn't works with the javascript engine used by React Native 
+  const date = new Date("10-10-2020 12:30");
+  ```
+  A good practice to get valid date could be:
+
+  ```javascript
+  // Get date to schedule a local notification today at 12:30:00
+  const hour = 12;
+  const minute = 30;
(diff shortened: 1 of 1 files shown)


Dallas62:
Hi @Scriptonita 
Thanks for this information. I don't fully understand the solution, can you elaborate here ?
When do you meet that issue ? with string date, or something else ?

Because you could use:

```js
const date = new Date();

date.setHours(12, 30, 00);

// OR "old-way"

date.setHours(12);
date.setMinutes(30);
date.setSeconds(00);
```

This looks simple but doesn't check is the date is in the past, if `new Date() > 12:30:00`.

https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Date/setHours

Scriptonita:
Hi @Dallas62, I will try to explain myself.

In my app, I was using a string like "08-26-2020 10:30"

```javascript
const date = new Date(`${month}-${day}-${year} ${hour}`);
```
It works fine while you are with the remote debugger enabled and you will get local notifications scheduled working
But It doesn't work when you have disabled remote debug.

For me was very hard to found that it was the problem because as I said, it works in debug mode and I refactorized a lot of code trying to find the problem.

This PR is only to help someone that could have the same problem. 
The piece of code is only a suggestion that works for me.

trouble-shooting.md
```diff
@@ -7,6 +7,25 @@ Known bugs and issues:
  * (Android) Tapping an alert in the notification centre will sometimes not result in `onNotification` being called [issue 281](https://github.com/zo0r/react-native-push-notification/issues/281)
  * (Android) Not all local notification features are supported yet (PRs welcome)
  * (iOS) The OS can penalise your app for not calling the completion handler and will stop (or delay) sending notifications to your app. This will be supported from RN-0.38 [PR 227](https://github.com/zo0r/react-native-push-notification/pull/277)
+ * (Android and iOS) If local notifications only works with remote debugger enabled it could be due to the date that you are adding to the notification, [explanation](https://stackoverflow.com/a/41881765/8519917).
```
Dallas62, line 10: Is it possible to precise that this issue occures when using string format ? it was not clear as first read 😉

Scriptonita: Changed!

PR #1686, opened 2020-10-04
Fix for vibration on notifs for Android API >= 26

mineshpatel1:
This fixes the vibration feature for notification for Android notifications for API >= 26. This pattern is used for the Notif Builder for API < 26 and it appears to work there as well.

Before:

```
<Sending Notif>
2020-10-04 12:37:03.630 1936-29464/? E/VibratorService: vibratorOff command failed (1).
2020-10-04 12:37:03.935 1936-29476/? E/VibratorService: vibratorOff command failed (1).
```

After:
```
<Sending Notif>
2020-10-04 12:07:16.037 1936-4760/? E/VibratorService: vibratorOff command failed (1).
2020-10-04 12:07:16.037 1936-4760/? E/VibratorService: vibratorOn command failed (1).
2020-10-04 12:07:16.438 1936-1936/? E/VibratorService: vibratorOff command failed (1).
```

Have tested this on a physical device as well and it works. This could lead to an obvious extension which is for users to specify a vibration pattern array when creating a channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,7 +406,7 @@ To use channels, create them at startup and pass the matching `channelId` throug
   );
 ```
 
-**NOTE: Without channel, remote notifications don't work**
+**NOTE: Without channel, notifications don't work**
 
 In the notifications options, you must provide a channel id with `channelId: "your-channel-id"`, if the channel doesn't exist the notification might not e triggered. Once the channel is created, the channel cannot be update. Make sure your `channelId` is different if you change these options. If you have created a channel in another way, it will apply options of the channel.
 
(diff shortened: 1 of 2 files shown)


Dallas62:
Thanks for the PR!
I will take a look at this ASAP!
Regards

mineshpatel1:
@Dallas62 - Bump, I think there's still value in landing this one.

mineshpatel1:
@Dallas62 - Hi, would you be able to review this pull request please? Just making sure it's not been forgotten about, it's just a little one (but an important fix for newer versions of Android).

Dallas62:
Hi,
I will check this PR ASAP (holidays)
Regards

juancabrera:
@mineshpatel1 thanks for the fix, made the change manually on my local and works as expected!

Dallas62:
Hi,
This line should probably be updated too:
https://github.com/zo0r/react-native-push-notification/blob/b89eb226d2279195368a891bddf0e825a6130616/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L449

Dallas62:
Forget my last comment/commit, reverted, was tired 😄 
You will find this in the 6.1.2, and I also test on real device and simulator.

You may encounter this bug in old version if the channel is already created.

Two options:
- Delete then recreate the channel.
- Change the channel id to ensure this is the correct settings (and remove the old one if needed)

cristianoccazinsp:
Heads up, this only seems to work if the channel is re-created / full app re-install.

Issue #1432, opened 2020-05-11
Android - No sound on notifications?


cristianoccazinsp:
# Bug

Using `{playSound: true, soundName: 'default'}` doesn't work with `localNotification`.  This affects both iOS and Android. Are there limitations on local notifications? I am able to turn on sound on Android if I configure the notification channel (from the OS settings) to use sound.

## Environment info

Library version: 3.4.0

UxmanKaxmi:
okay so this might sound a little stupid but it might be an issue for you too.

I was using the whole string "android.resource://com.xyz/raw/my_sound" instead of just using "my_sound".  :D 
You just have to use the exact name, not the whole string. I think someone should look at the comment which is written in the latest commit.

cristianoccazinsp:
I'm using just `'default'` (no custom sounds), yet the notification is set as no sound. What version are you using?

Probably related:
https://github.com/zo0r/react-native-push-notification/issues/1143
https://github.com/zo0r/react-native-push-notification/issues/1045

UxmanKaxmi:
Oh. My default sound was playing fine.  My version is
`    "react-native-push-notification": "^3.4.0",
`

`    "react-native": "0.62.2",
`

and I am testing on my Samsung Note 8.

cristianoccazinsp:
I've tested on a Pixel 2 (Android 10) and Motorola G5 (Android 8.1), neither seem to work.  It seems like the notification channel is created muted as default.

UxmanKaxmi:
Can you show the code?

cristianoccazinsp:
Sample Code:

```
import PushNotification from 'react-native-push-notification';


PushNotification.configure({

  // (optional) Called when Token is generated (iOS and Android)
  // NOT using push notifications from the server right now, this step is not done.
  onRegister: function(token) {
  },

  // (required) Called when a remote or local notification is opened or received
  onNotification: function(notification) {

    // required on iOS only
    if(Platform.OS == 'ios'){
      notification.finish(PushNotificationIOS.FetchResult.NewData);
    }
  },


  // ANDROID ONLY: GCM or FCM Sender ID (product_number) (optional - not required for local notifications, but is need to receive remote push notifications)
  //senderID: "YOUR GCM (OR FCM) SENDER ID",

  // IOS ONLY (optional): default: all - Permissions to register.
  permissions: {
    alert: true,
    badge: true,
    sound: true
  },

  // Should the initial notification be popped automatically
  // default: true
[...]

Issue #1045, opened 2019-04-13
mute sound for local notification not working


roysG:
I tried to change the value playSound to "false" but the notification sound is still on.
What should i do?

cristianoccazinsp:
Facing the same issue. Local notifications always have sound AND vibration; can't mute them. Any thoughts?

cristianoccazinsp:
Most likely related to the importance/priority flags. Since priority was deprecated, and notification channel importance is now used, based on the notification channel importance the notification can or not play sounds. This is from testing a notification channel with a foreground service. However, I wasn't able to mute the notification even with a low importance/priority.

Shoaib-Raza:
Any update on this issue.
I am also facing the same issue, i want to mute the sound but playSound = false doesn't work.

necto68:
https://github.com/zo0r/react-native-push-notification/issues/1140#issuecomment-528331545

greenais:
The same issue here.
I expect my users to be disturbed by such inflexibility(

dgreasi:
Facing same issue here.

Versions:
"react-native-push-notification": "^6.1.2",
"react-native": "0.63.2",

Any updates?

dgreasi:
I actually found a workaround that is working on android but it will be also a silent notification, which means it will not be visible if you dont open your notifications tray.

In my case is what i wanted

On the createChannel you need to specify a low importance. Thus the working code for me:
```
PushNotification.createChannel(
    {
      channelId: 'silent-notification',
      channelName: 'A silent notification channel,
      importance: 2, // < 3 seems to do the job
      vibrate: false,
      soundName: undefined, // works even if it is not specified
    },
    (created) => console.log(`createChannel returned '${created}'`),
  );
```

The function that i use to trigger my local notification:
```
export const showNotification = (
  title: string,
  message: string,
  isSticky: boolean = false,
  soundName = 'default',
  isSilent = true,
) => {
  PushNotification.localNotification({
    title,
    message,
    channelId: 'silent-notification',
    ongoing: isSticky,
    vibrate: !isSilent,
    playSound: !isSilent,
    soundName,
    smallIcon: 'ic_notification_video',
    largeIcon: 'bootsplash_logo',
  });
};
```

sonnd08:
The same issue

PR #1736, opened 2020-11-08
Android: silent channel using playSound flag

halaei:
Currently it isn't possible to make silent channels, and sending silent notifications to Android > 8 devices.
A playSound option is added to PushNotification.createChannel() to make it possible to create silent channels in Android. The same option exists in PushNotification.localNotification().

Related issues: https://github.com/zo0r/react-native-push-notification/issues/1432 and https://github.com/zo0r/react-native-push-notification/issues/1045

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,7 @@ To use channels, create them at startup and pass the matching `channelId` throug
       channelId: "channel-id", // (required)
       channelName: "My channel", // (required)
       channelDescription: "A channel to categorise your notifications", // (optional) default: undefined.
+      playSound: false, // (optional) default: true
       soundName: "default", // (optional) See `soundName` parameter of `localNotification` function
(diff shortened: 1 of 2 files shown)


Dallas62:
Hi,
What about setting the sound as an empty string ?
Regards

halaei:
🤷🏻‍♂️ Empty string probably works fine. If backward compatibility isn't a concern, I would like to use anything like null/undefined/false/"silente"/"none" too. I am not sure which option is more relatable.
As I mentioned, I thought maybe because there already is a playSound option in the other place, it's OK to have it here too.

Dallas62:
👍 
Is it possible to reflect the change in the README ?

halaei:
Sure. In a day.

Dallas62 (approved):

PR #1740, opened 2020-11-10
docs: use package import

jeremybarbet:
Hey!

I believe it's should be imported from the repository's library and not using `PushNotificationIOS`.


Dallas62 (approved):

Dallas62:
Thanks @jeremybarbet !

PR #1760, opened 2020-11-26
fixed duplicate sending of notifications in the foreground

Dimon70007:


android/src/main/java/com/dieam/reactnativepushnotification/modules/RNReceivedMessageHandler.java
```diff
@@ -172,7 +172,7 @@ private void handleRemotePushNotification(ReactApplicationContext context, Bundl
             jsDelivery.notifyRemoteFetch(bundle);
         }
 
-        if (config.getNotificationForeground() || !isForeground) {
+        if (config.getNotificationForeground() && !isForeground) {
```
Dallas62, line 175: This condition doesn't make sense:
If "Allow notifications in foreground" and "Not in foreground":
Then Trigger the notification.

you probably missed a configuration on the AndroidManifest or in your app configuration...

Dimon70007: With those changes showing local notifications works as expected (without duplication)
I use this repo in combine with react-native-firebase v6 messaging

Dallas62: I'm not saying this change is not working for you, I'm saying that this change doesn't have sense (see the comment above).
If you experience duplicate notifications, this means there is a configuration issue in your project.

Dimon70007: > NOTE: localNotification() works without changes in the application part
this is from readme, i use local notifications only with this library.

Dimon70007: anyway, thank you for great library

Dallas62:
```xml
  <!-- Change the value to true to enable pop-up for in foreground (remote-only, for local use ignoreInForeground) -->
        <meta-data  android:name="com.dieam.reactnativepushnotification.notification_foreground"
                    android:value="false"/>
```

Dimon70007: Thanks, i've created a new PR about readme updating - https://github.com/zo0r/react-native-push-notification/pull/1762

PR #1761, opened 2020-11-27
Update peerDependency `push-notification-ios`  to `^1.7.3`

ulayab:
I faced a bug about `userInteraction` variable, seems had fixed in version 1.7.3


Dallas62:
Hi @ulayab 
Thanks for this PR, this change is already reflected into https://github.com/zo0r/react-native-push-notification/pull/1751/files
😉 
Regards,

PR #1793, opened 2020-12-26
(fix) deep clone details and notifications

Stanislavja:
should not modify external variables. add deep clone.


Dallas62:
Hi @Stanislavja 
Released in 7.2.2.
Regards,

farwayer:
This is not deep clone. There are a several places where user data is still modified:
https://github.com/zo0r/react-native-push-notification/blob/624dd152fa06471e16d3e157aa4f058fe0a01467/index.js#L239
https://github.com/zo0r/react-native-push-notification/blob/624dd152fa06471e16d3e157aa4f058fe0a01467/index.js#L159

Issue #1801, opened 2021-01-01
Android: Handling received push notifications (in app's foreground/background state)


robertgreso:
According to docs, `onNotification` is "Called when a remote is received or opened". I would like to ask, if its possible to somehow get the behaviour that onNotification would be on Android called once push notification is opened, not just once its received (when Android app is in foreground/background)?

When implemented on iOS, onNotification is always called just when I click on received push notification, not immediately once its received. I would like to get the same behaviour on Android as well but it behaves differently there.
On Android, if the app is in the foreground/background `onNotification` is called immediately once the remote notification is received, not when user actually clicks on received notification.

Is it possible to get the same behaviour as described above on iOS also for Android, ie to be able to handle received push notification on Android while the app is in foreground/background just when user opens received push notification, not just when the notification is received?

Thanks for any suggestions!

Dallas62:
Hi @robertgreso 

This question has already been answered in the last ten issues.

Regards,

robertgreso:
> Hi @robertgreso
> 
> This question has already been answered in the last ten issues.
> 
> Regards,

Hi @Dallas62, I checked all the issues before creating this one and also double checked now but couldnt find the answer how to make it work to get onNotification called on Android once user taps on notification, when in foreground/background.

I just found as you pointed someone else to troubleshooting page where it says "Tapping the alert in the notification centre will start or bring the app to the foreground and onNotification will be called.", so I guess onNotification should be called in the case I described but it is not working for me. 
Could you please help me here? Much appreciate it! Thank you!

Dallas62:
https://github.com/zo0r/react-native-push-notification/issues/1795#issuecomment-752380816

robertgreso:
Hi @Dallas62 !
 Thanks for quick response!  
 I checked that comment and based on it:  We use the newest 7.0.0 version and we call .configure() outside of any component.
  
    I searched for SplashScreen in the issues (https://github.com/zo0r/react-native-push-notification/issues?q=is%3Aissue+splashscreen).  The relevant issues I found which are about push notifications:  
  
  https://github.com/zo0r/react-native-push-notification/issues/1795
	- this one is just about killed state where I have issue with foreground/backgroud state, killed state works fine for me

https://github.com/zo0r/react-native-push-notification/issues/1592
	- we dont use react native splash screen in our app

https://github.com/zo0r/react-native-push-notification/issues/868
	- we do send only data key in payload so that should be fine

[...]

PR #1819, opened 2021-01-13
[Android] Add hooks to intent handling and bundle parsing

cayleyh:
For https://github.com/zo0r/react-native-push-notification/issues/1801.

This adds hooks to allow easier customization of intent handling and bundle parsing behaviour on Android. This can be used to send notifications sent by 3rd party messaging management platforms that have custom message formats the same onNotification() JS codepath.

See discussion in #1801.

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,6 +702,49 @@ Uses the [ShortcutBadger](https://github.com/leolin310148/ShortcutBadger) on And
 
 `PushNotification.unsubscribeFromTopic(topic: string)` Unsubscribe from a topic (works only with Firebase)
 
+## Android Custom Notification Handling
+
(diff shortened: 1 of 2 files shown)


Dallas62:
Hi @cayleyh 

I will look into it ASAP.

Regards,

Dallas62:
Hi @cayleyh 

Can you provide an exemple in the `README.md` ? Like the one in #1801 ?
Even if there is already many things inside the README and it needs refactor 😄. Because if this feature is not documented, this will not be reusable.

The batch exemple looks enough. A release will be made this week-end.

Regards,

cayleyh:
For sure. I'll add an example to the README right away.

cayleyh:
@Dallas62 I've added instructions to the README on the 2 separate cases we discussed in #1801.

Dallas62 (approved):

PR #1831, opened 2021-01-19
unsubscribeFromTopic function fix

MostafaArafa95:
Please note that this merge https://github.com/zo0r/react-native-push-notification/pull/917 doesn't work 

This is a fix to the unsubscribeFromTopic function


Dallas62:
Nice catch !

PR #1847, opened 2021-02-03
Remove redundant import statement

vaibhavhrt:
same modules were imported twice.


Dallas62:
Hi @vaibhavhrt 
Thanks for the PR ! 😉 
I will release it soon.

PR #1905, opened 2021-03-10
(fix) remove unnecessary imports and extra variables.

Stanislavja:
Remove extra variables.
Led to one look all imports.


Dallas62:
Hi @Stanislavja 

Thanks for this change !

PR #1910, opened 2021-03-13
Document `data` in getScheduledLocalNotifications

s-fletcher:
I noticed that `userInfo`/`data` was being returned in the notifications array of the `getScheduledLocalNotifications` but was not properly documented. 

See https://github.com/zo0r/react-native-push-notification/blob/master/index.js#L546

I've also added this property to the @types package https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51720


Dallas62:
Hi @s-fletcher 
Thanks for this PR

PR #1956, opened 2021-04-18
Adding mavenCentral() as jcenter() is shutting

maheshwarimrinal:
Adding mavenCentral() as jcenter() is shutting down


Dallas62:
Hi @maheshwarimrinal 
Thanks for you contribution !

PR #1995, opened 2021-05-12
fix: android missing channelId warning should now show

gioragutt:
Small one, would have been caught by TypeScript easily 🤷🏻‍♂️

### DIFF
--- a/index.js
+++ b/index.js
@@ -141,7 +141,7 @@ Notifications.unregister = function() {
  * @param {Object}    details.userInfo -  iOS ONLY: The userInfo used in the notification alert.
  */
 Notifications.localNotification = function({...details}) {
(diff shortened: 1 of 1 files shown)


Dallas62:
Thanks !

Dallas62:
Will release it ASAP

Aarush-wadhia:
Hi @darryl-tan this warning is usually given in android as referring to create a channel for triggering the notification. As of Android O, creating a Notification instance requires a channel ID. This is to ensure that our notification belongs to a channel that can be managed by the user from their device settings. And as of solution for your problem is this:
```
const showNotification = (title, message, channelId, channelName) => {
  PushNotification.createChannel({
            channelId,
            channelName,
        });

          channelId,
          title:title,
          message:message,
        });
}
```

PR #2027, opened 2021-05-24
Pull Latest from zo0r / react-native-push-notification to explosiv0/react-native-push-notification

vinoth-hcl:
Firebase library instance ID deprecation fix.. explosiv0/react-native-push-notification


vinoth-hcl:
Please pull latest code from zo0r/react-native-push-notification as firebase deprecation fixed there in 7.3.1

PR #2030, opened 2021-05-25
Feat: Allow for repeat to specify amount of the given repeat type (Android)

johgusta:
# Problem
We want our local notifications to be able to repeat every other day, and it is critical that the time of day is correct. Currently the closest possible setting would be to use `repeatType: 'time'` and then calculate the milliseconds for two days and set it as `repeatTime`. For a repeating reminder this will change the time of day the reminder repeats when daylight saving changes the UTC offset. 

# Solution
Since the library already allows `repeatTime` for `time`, this PR will update it to allow `repeatTime` for all the repeat types. Instead of calculating the milliseconds of each type we're using the `Calendar.add()` function with the corresponding field. The amount to increase it by is set by `repeatTime` and defauls to 1. It also follows the same logic as already implemented for repeating on months.

Example of how to repeat every other day:
```javascript
PushNotification.localNotificationSchedule({
    ...
    repeatType: 'day',
    repeatTime: 2,
    ...
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,7 @@ PushNotification.localNotification({
   ignoreInForeground: false, // (optional) if true, the notification will not be visible when the app is in the foreground (useful for parity with how iOS notifications appear). should be used in combine with `com.dieam.reactnativepushnotification.notification_foreground` setting
   shortcutId: "shortcut-id", // (optional) If this notification is duplicative of a Launcher shortcut, sets the id of the shortcut, in case the Launcher wants to hide the shortcut, default undefined
   onlyAlertOnce: false, // (optional) alert will open only once with sound and notify, default: false
+  repeatTime: 1, // (optional) Increment of configured repeateType. Check 'Repeating Notifications' section for more info.
   
   when: null, // (optional) Add a timestamp (Unix timestamp value in milliseconds) pertaining to the notification (usually the time the event occurred). For apps targeting Build.VERSION_CODES.N and above, this time is not shown anymore by default and must be opted into by using `showWhen`, default: null.
   usesChronometer: false, // (optional) Show the `when` field as a stopwatch. Instead of presenting `when` as a timestamp, the notification will show an automatically updating display of the minutes and seconds since when. Useful when showing an elapsed time (like an ongoing phone call), default: false.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -324,6 +324,7 @@ PushNotification.localNotification({
   ignoreInForeground: false, // (optional) if true, the notification will not be visible when the app is in the foreground (useful for parity with how iOS notifications appear). should be used in combine with `com.dieam.reactnativepushnotification.notification_foreground` setting
   shortcutId: "shortcut-id", // (optional) If this notification is duplicative of a Launcher shortcut, sets the id of the shortcut, in case the Launcher wants to hide the shortcut, default undefined
   onlyAlertOnce: false, // (optional) alert will open only once with sound and notify, default: false
+  repeatTime: 1, // (optional) Increment of configured repeateType. Check 'Repeating Notifications' section for more info.
```
Dallas62, line 327: Hi @johgusta 
This line should not be documented there, it's on the scheduldedNotification method 😉

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
```diff
@@ -673,6 +643,22 @@ private void scheduleNextNotificationIfRepeating(Bundle bundle) {
...
     }
 
+    private int getRepeatField(String repeatType) {
+        switch (repeatType) {
+            case "month":
+                return Calendar.MONTH;
+            case "week":
+                return Calendar.WEEK_OF_YEAR;
+            case "day":
+            default:
```
Dallas62, line 653: Is it possible to move the `default` at the bottom of the `switch` ?

Dallas62:
Hi @johgusta 
Thanks for the PR, can you take a look at comments ?
I will test the PR just after changes 😉 
Regards

README.md
johgusta: I might have misunderstood this, but I think repeating is possible for both the `localNotification` and `localNotificationSchedule` functions. Since this one has the documentation for `repeatType: 'day'` I thought this would be the right place to add this info. In my project we only use `localNotificationSchedule` though but as far as I can tell it should still work for both.

android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
johgusta: Sure! I've moved it to the bottom now.

johgusta:
> Hi @johgusta
> Thanks for the PR, can you take a look at comments ?
> I will test the PR just after changes 😉
> Regards

Thanks for the review! I've updated the code and I think the documentation is correct as it is, but let me know if I've misunderstood that and I'll fix it.

Dallas62:
Hi @johgusta 
Thanks for the update, the issue with the documentation is that there is `repeatTime` added at a wrong place for `localNotification`, this option is not available for this method

johgusta:
> Hi @johgusta
> Thanks for the update, the issue with the documentation is that there is `repeatTime` added at a wrong place for `localNotification`, this option is not available for this method

Ok! I will update the documentation and move it to the the scheduled notification section then.

Dallas62:
Will take a look ASAP and release it.

PR #2063, opened 2021-06-24
feature(ios): Add support for subtitle notification property

GertSallaerts:
Add support for `subtitle` on iOS as seen in https://github.com/react-native-push-notification-ios/push-notification-ios#addnotificationrequest


Dallas62:
Hi @GertSallaerts 
Thanks for this PR !
We should probably reuse the property `subText` of Android since it's the same purpose. 
Or change `subText` to `subtitle`.

GertSallaerts:
Hey @Dallas62, thanks. That's what I wanted to do at first but I tested `subText` on Android and wasn't sure if it should be the same as `subText` seems a bit less "important" then iOS' subtitle. But I'll leave that up to you, let me know what you would prefer.

As a visual aid I pulled two examples from the web:

| Android subText | iOS subtitle |
| -- | -- |
| image | image |

Dallas62:
One day they will probably make something uniform ... 😄

Dallas62:
Thanks for your contributions ! 😃

GertSallaerts:
You're welcome, thanks for the quick merge and release :100:

PR #1469, opened 2020-06-07
[ANDROID] support html tags for message and bigText fields

wanderbon:
Support popular html tags in android notifications (without img)

PR #2067, opened 2021-06-26
Allow HTML tags to be used for styling in bigText field only

silencer07:
**Rationale:**

bigText field should be allowed to have HTML tags as a way to format the message since this has been already supported in android natively for a long time

**Expected behaviour:**

Allows a subset of html tags to be used as outlined [in this android documentation](https://developer.android.com/guide/topics/resources/string-resource#StylingWithHTML). Note that if user did not provided a bigText field but added html tags in message it self, it won't work as message field in iOS does not support HTML tags nor iOS has bigText property

**Testing:**

on bigText field, add some `<br/>` tags and see if it will put the subsequent string to the next line in android.

**Change:**

This is a small change, this merely decorates the bigText as Spanned Text

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ PushNotification.localNotification({
   largeIcon: "ic_launcher", // (optional) default: "ic_launcher". Use "" for no large icon.
   largeIconUrl: "https://www.example.tld/picture.jpg", // (optional) default: undefined
   smallIcon: "ic_notification", // (optional) default: "ic_notification" with fallback for "ic_launcher". Use "" for default small icon.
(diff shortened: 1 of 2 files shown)


silencer07:
This PR is a modified and simplifed version of [wanderbon's PR](https://github.com/zo0r/react-native-push-notification/pull/1469) that is not yet merged

Dallas62:
Hi,
Thanks for the PR, will not be available this month, will look when I comme back.
Regards

silencer07:
@Dallas62 hello?

Dallas62:
Hi @silencer07 
Regards

silencer07:
Thanks @Dallas62

Issue #2100, opened 2021-07-21
cancelLocalNotifications() does not work on Android


jamesxabregas:
I have tested this extensively to see if there was something wrong with the way I was excuting cancelLocalNotification() but it does not appear that this works when providing an id on Android. I am executing it exactly per the documentation.

After adding debug logging into the Anrdoid pacakge I believe I have found the root cause of the issue. In `RNPushNotificationHelper.java` the method `cancelScheduledNotification(ReadableMap userInfo)` tries to extract the id from the provided JSON map and then find the correct notification to cancel. 

https://github.com/zo0r/react-native-push-notification/blob/6a81fba73b55f1ac975a2568486297102998150c/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L783-L786

The problem here is that the `matches` method in `RNPushNotificationAttributes.java` is being used incorrectly in this case. That method tries to match every element of the userInfo object with the notification it is iterating on which contains additional info such as the title, message etc. If the user has only provided the id attribute this always returns false.

https://github.com/zo0r/react-native-push-notification/blob/6a81fba73b55f1ac975a2568486297102998150c/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationAttributes.java#L200-L247

What is needed in this check is a simple method that matches the id's only. 

For instance something like this in `RNPushNotificationHelper.java`:
```
RNPushNotificationAttributes notificationAttributes = fromJson(notificationAttributesJson);
if (notificationAttributes.idMatches(userInfo)) {
    cancelScheduledNotification(id);
}
```

and this in `RNPushNotificationAttributes.java`

```
public boolean idMatches(ReadableMap userInfo) {
try {
    if(this.userInfo == null || userInfo == null) {
        return false;
    }
        return this.id.equals(userInfo.getString("id"));
    } catch(Exception e) {
        return false;
    }
}
```

I am happy to submit a pull request for this issue. I am working on my own fork at the moment.

Dallas62:
Hi @jamesxabregas 
Did you cross check cases on iOS?
With multiple data in userInfo?

There is another method on Android which allow remove a scheduled notification only with the id.

jamesxabregas:
I've tested this on iOS and it is not an issue. The issue only occurs on Android, and as explained it is due to an error in the Android implementation. I have also tested this with other data in the userInfo object but that doesn't make a difference.

[...]

PR #2101, opened 2021-07-21
Fix bug #2100 in cancelLocalNotifications() for Android

jamesxabregas:
cancelLocalNotifications() does not work on Android due to an implementation error. This PR fixes the issue #2100.


garrettmaring:
It would be great to see this go in (assuming that it fixes the Android side) as I'm also running into this problem. @jamesxabregas you might want to remove the formatting changes, which seem to be deleting tab indents, that're added in this PR as they clutter the diff.

Very much appreciated that you found a fix for our issue 🥂

Dallas62:
Fixed in 8.0.0.
Regards

garrettmaring:
Much obliged @Dallas62!

Dallas62:
It's basicly the same king of fix, I just remove extra steps since there was a private method to acheive it. iOS doesn't use the `userInfo` to cancel notification in recent versions.

PR #2125, opened 2021-08-24
Fix 2122

dimadeveatii:
Fixes canceling notifications on Android.

See:  https://github.com/zo0r/react-native-push-notification/issues/2122


Dallas62:
Thanks for the PR !

Issue #2129, opened 2021-08-30
Android 12: PendingIntent for notifications will not work


cristianoccazinsp:
# Bug

## Environment info

`react-native info` output:

```bash
 System:
    OS: macOS 11.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 48.70 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.22.1 - /usr/local/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.12 - /usr/local/bin/npm
    Watchman: 2021.06.07.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK:
      API Levels: 28, 29, 30, 31
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 31.0.0
      Android NDK: Not Found
  IDEs:
    Android Studio: 4.1 AI-201.8743.12.41.7199119
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_121 - /usr/bin/javac
    Python: 2.7.15 - /Library/Frameworks/Python.framework/Versions/2.7/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.14.0 => 16.14.0 
    react-native: 0.62.2 => 0.62.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

Library version: 8.0.1

## Steps To Reproduce

Running: `PushNotification.localNotification({...})` will crash in Android 12.

Due to Android 12 changes, trying to display a local or remote notification will fail with the given error on Android 12 (when compiling with SDK 31):

```
2021-08-30 14:23:16.910 29950-30022/com.zinspector3.dev E/RNPushNotification: failed to send push notification
    java.lang.IllegalArgumentException: com.zinspector3.dev: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
[...]

PR #2130, opened 2021-08-30
Fix Android 12 PendingIntent

cristianoccazinsp:
Fix Android 12 (SDK 31) notifications due to missing PendingIntent flags.

Fixes https://github.com/zo0r/react-native-push-notification/issues/2129

I've tested this on a Pixel 5 with Android 12 and a RN app compiled with SDK 31. However, I've only tested local notifications, but I expect that the change should work across all notifications as it is very simple.

### DIFF
--- a/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
+++ b/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java
@@ -107,7 +107,7 @@ public class RNPushNotificationHelper {
             notificationIntent.putExtra(RNPushNotificationPublisher.NOTIFICATION_ID, notificationID);
             notificationIntent.putExtras(bundle);
 
-            return PendingIntent.getBroadcast(context, notificationID, notificationIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+            return PendingIntent.getBroadcast(context, notificationID, notificationIntent, Build.VERSION.SDK_INT >= Build.VERSION_CODES.M ? PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE : PendingIntent.FLAG_UPDATE_CURRENT);
         } catch (Exception e) {
             Log.e(LOG_TAG, "Unable to parse Notification ID", e);
         }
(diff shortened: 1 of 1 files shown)


Dallas62:
Hi @cristianoccazinsp 
Thanks for your contributions !

cristianoccazinsp:
Thanks for merging, I think I missed this line too: https://github.com/cristianoccazinsp/react-native-push-notification/blob/8d7ab959f574984bac7df50f5adb84bdb5e04b88/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L531

Can you update that line with the same logic before releasing?

Dallas62:
Hi @cristianoccazinsp
This change affect some functionalities: #2152 
Since I don't have time to investigate, do you have an idea ?

cristianoccazinsp:
@Dallas62 I think the issue may be related to using `PendingIntent.FLAG_IMMUTABLE` instead of `FLAG_MUTABLE`.  I don't currently have code that uses #2152 code, but perhaps @redexp or @TongManhCong can help testing the change.

Dallas62:
@redexp @TongManhCong

You can try to change :
https://github.com/zo0r/react-native-push-notification/blob/master/android/src/main/java/com/dieam/reactnativepushnotification/modules/RNPushNotificationHelper.java#L536

by:
```java
                    int flags = PendingIntent.FLAG_UPDATE_CURRENT;

                    if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
                      if(action.equals("ReplyInput")) {
                        flags = PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_MUTABLE;
                      }
                      else {
                        flags = PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE;
                      }
                    }
```
Regards

cristianoccazinsp:
That seems like an even more reasonable fix, checking for reply input as well. 

What are the side effects of marking an intent as mutable?

ammarahm-ed:
The current fix has broken ReplyInput. `reply_text` is not returned by `onAction` because `PendingIntent.FLAG_IMMUTABLE` is set.

Issue #2147, opened 2021-09-13
Specific time in week repeatType


ilyasbat:
Is that possible? 

```
 PushNotification.localNotificationSchedule({
      date: date,

      channelId: ANDROID_CHANNEL_NAME,
      id: id, // (optional) Valid unique 32 bit integer specified as string. default: Autogenerated Unique ID
      vibrate: true,
      vibration: 300,
      title: title, // (optional)
      message: message, // (required)
      userInfo: {screen: 'blabla'},
      soundName: 'bell.mp3', 
      repeatType: 'week',
      allowWhileIdle: true,
    });
```
I use these codes but the time is not set. 00:00 appears as the notification time. Can't I set weekly notifications at different times on the same day?

Dallas62:
Hi,

The time of the notification is the one defined in `date`. It cannot be different on each repetition.

Regards

ilyasbat:
Thanks. I decide to schedule again after every notification.

Dallas62:
Take care of the number of notification you define, in iOS it's 64, and on Samsung too if I remember correctly.
If this exceed, notifications will not trigger (even if set sooner than other) or it will never trigger a notification (depend on manufacturers).

Dallas62:
Take care of the number of notification you define, in iOS it's 64, and on Samsung too if I remember correctly.
If this exceed, notifications will not trigger (even if set sooner than other) or it will never trigger a notification (depend on manufacturers).

ilyasbat:
A scheduled notification is canceled automatically. If an old notification was not canceled. I cancel so i think no problem in this situation. Totally 40 notifications set.

cynde:
> Thanks. I decide to schedule again after every notification.

Hi how do you schedule again after every notification? Because I need to do it too but cannot find a way since onNotification() in index.js only called if a notification is clicked.

zeeshan112233:
Hi, @cynde  did you manage to figure this out? I also need to do something on a notification received without clicking it.

cynde:
@zeeshan112233 hi. still don't find it. I managed to use other way without using react-native-push-notification's feature.

anhquan291:
> @zeeshan112233 hi. still don't find it. I managed to use other way without using react-native-push-notification's feature.

Hi @cynde. Could you please share the way you handle this issue? Really appreciate that. Thanks

PR #2150, opened 2021-09-21
Fix iOS repeated notification timing

mmmoussa:
For repetition over larger time intervals, the smaller time intervals within `repeatsComponent` need to be set to `true` in order to ensure the notification is delivered at the correct time. Without this fix, iOS scheduled repeating notifications are not sent correctly and do not match the Android behaviour. 

I have tested this PR on my own app in prod and it fixes the issue correctly.

#2147 may be related.

### DIFF
--- a/index.js
+++ b/index.js
@@ -252,9 +252,9 @@ Notifications.localNotificationSchedule = function({...details}) {
     }
     
     const repeatsComponent = {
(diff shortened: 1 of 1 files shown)


MCElmo:
+1 This fixed the issue for me

mohsenV1980 (approved):
Exactly.

PR #2159, opened 2021-09-23
Updated Firebase reference to show up in English

espi-jisc:
😃


Dallas62:

PR #2199, opened 2021-11-18
Expose `onNotification` event listener via new api `onNotificationOpenedApp`

sanduluca:
This new api was inspired from `@react-native-firebase/messaging`.

I was struggling trying to configure navigation to a particular screen when a notification was pressed. All the examples I found does not provide full freedom or good explanation on how this can be configured. I was inspired from a post I found on medium.com (https://medium.com/cybermonkey/deep-linking-push-notifications-with-react-navigation-5fce260ccca2). I found it one day and I was wondering why can't this be done with `react-native-push-notifications`.

**An example with usage can be found in the README.** 

**Also an example of integration with react-navigation via deep linking was added.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,20 @@ Notification object example:
     data: {}, // OBJECT: The push data or the defined userInfo in local notifications
 }
 ```
+```js
+PushNotification.onNotificationOpenedApp(handler: (notification: Object) => EmitterSubscription)
+```
+
+EXAMPLE:
+
+```js
+useEffect(() => {
+  const subscription = PushNotification.onNotificationOpenedApp((notification) => {
+    console.log('Notification was opened', notification)
+  })
+  return () => subscription.remove()
+}, [])
+```
 
 ## Local Notifications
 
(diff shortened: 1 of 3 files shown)


Dallas62:

sanduluca:
> https://github.com/zo0r/react-native-push-notification#state-of-the-repository

In response to the State of the repository
https://www.npmtrends.com/@notifee/react-native-vs-react-native-push-notification-vs-react-native-notifications 😀

Dallas62:
I like your answer, but this is still not giving time for maintenance 😉 
I will not spend more time on this project, too many refactor needed.
Your new feature is cool, but it's also creating a new way of getting a notification.

There is already a lot of mix in the code with Promises, Callbacks and now the new event emitter...

I don't have time for that, even with a xxxB dowloads/week, this is not giving me time nor pay my bills.

I take the maintenance of this project to fix lake of functionalities/issues. Since september notifee which provide a full feature library on both platform and have some people payed to maintain this library. I already switch my apps to this one.

If you want to take maintainership of this repository, feel free to contact @zo0r 

Regards

PR #2229, opened 2022-02-02
Fix optional parameters for repeatsComponent

BrandonMA:
Daily notifications are not currently working because we set `day: false` instead of not passing a `day` value at all.


Dallas62:

PR #2230, opened 2022-02-02
Use root locale when changing case in java

halaei:
Not setting locale for language/country neutral operation may cause bug depending on the default locale.
See https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#ROOT

Note: I am just searching for toLowerCase() and toUppercase() in my project's dependencies and send the same PR, in order to potentially raise the awareness. Although I've seen the lack of explicit locale has caused issues for us, I am not sure if react-native-push-notification is actually affected.

Example related issue: joltup/rn-fetch-blob#573


Dallas62:

PR #2242, opened 2022-03-11
Fixed repeatType for day

ktarasenkowow:
Dont work date option if set repeatType="day"

### DIFF
--- a/index.js
+++ b/index.js
@@ -255,7 +255,7 @@ Notifications.localNotificationSchedule = function({...details}) {
       second: ['minute', 'hour', 'day', 'week', 'month'].includes(details.repeatType),
       minute: ['hour', 'day', 'week', 'month'].includes(details.repeatType),
       hour: ['day', 'week', 'month'].includes(details.repeatType),
-      day: details.repeatType == "month",
+      day: details.repeatType == "day",
       dayOfWeek: details.repeatType == "week",
     };
 


BrandonMA:
I think the right solution is this: https://github.com/zo0r/react-native-push-notification/pull/2229, maybe could merge this? It's a minimal change @Dallas62

ktarasenkowow:
@BrandonMA It is fixed dev build but broke release build... If u want test daily notification - change node_modules but stay original for release=) I dont now why repeatType === 'month' correct work for release-_-

BrandonMA:
Mmm weird, we have been using this fix at my company and it works. If you check the iOS documentation if you want the notification to repeat every day then you need `second`, `minute`, and `hour`. If you pass `day` it will repeat the notification every month, let's say I create a notification for April 1st and pass `day`, then the next notification I will receive is for May 1st.

In any case, you can also fork the project and consume the library directly from Github, that's what we did.

Dallas62:
As mentioned in the linked PR: https://github.com/zo0r/react-native-push-notification#state-of-the-repository

I highly recommend to move from this library to an other which is actively maintained.
I don't have time anymore, and I don't use the library anymore.

Regards.

Dallas62:
Maybe @Narmical will be able to help you soon.

PR #2322, opened 2022-10-31
substitute jcenter for  mavenCentral

nios87:
Since Jcenter is deprecated and having issues it might be good to use mavenCentral instead. Dont know how to test if this is a fix at the moment or if it works with this projects dependencies. Never done a PR on a package like this before.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,7 +2,6 @@ buildscript {
     repositories {
         mavenCentral()
         google()
-        jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.2.0'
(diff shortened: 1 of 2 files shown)


Baloo1:
@zo0r  @Dallas62 is it possible to get a review on this? 🙏

Dallas62:
@Baloo1 please read the Readme, this repository is no more maintained.

Baloo1:
@Dallas62 thanks for the reply I understand! Could you archive the repository then? So there is a super clear banner about it being unmaintained when you visit the repository.

Dallas62:
I do not have access to repository settings, initialy, for maintenance purpose and npm version.

Time is missing, debt is too high and better alternatives are available 🤷‍♂️

Baloo1:
@zo0r could you archive the repository then if its unmaintained? For others clarity

PR #2369, opened 2023-06-16
feat: add method to get fcm token

yousrasd:
- Implement a method to retrieve the FCM token upon request. I require this functionality in my app as I only need to obtain the FCM token once when the app is opened. I prefer not to use the onRegister method as it triggers continuously.

Goes with the following PR in [react-native-push-notification/ios](https://github.com/react-native-push-notification/ios): https://github.com/react-native-push-notification/ios/pull/410


jedabero:
Why not use `onRegister` passed to the configuration options?
```js
PushNotification.configure({
    requestPermissions: true,
    onRegister: ({ os, token }) => {
      // use token
    },
}):
```

PR #2373, opened 2023-07-05
Remove trackers

ororsatti:
For us, privacy focused fellas, removing any instance of trackers from the code, is very important. Thus, I'm creating this branch.
Note that this branch is not intended to get merged, since this lib allows for firebase analytics to be used, but since some of us don't want that, heres the alternative. 

If you have any comments on how to improve, change or add things, let me know.
I will try to maintain this branch, but as the lib it self is not maintained, I dont think I need to too much.

And lastly, a proof from exodus-standalone:
Screenshot 2023-07-05 at 17 27 22

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -60,5 +60,9 @@ dependencies {
     implementation "$appCompatLibName:$supportLibVersion"
     implementation 'com.facebook.react:react-native:+'
     implementation 'me.leolin:ShortcutBadger:1.1.22@aar'
-    implementation "com.google.firebase:firebase-messaging:${safeExtGet('firebaseMessagingVersion', '21.1.0')}"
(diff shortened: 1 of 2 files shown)


17Amir17:
Privacy first!

ksetrin:
@Dallas62 
Please have a loot this PR
We still use this lib with ❤️‍🔥

ramezrafla:
Can't we just set in firebase.json to disable tracking?

ororsatti:
> Can't we just set in firebase.json to disable tracking?

In our case we didn't even set up firebase, so... We didn't want it. 
Also, setting up firebase is (I assume) only disabling stuff, not removing them completely.
Exodus finds instances of the code itself .

PR #2394, opened 2023-11-28
removing FCM

bahaaTuffaha:
removing google and FCM


bahaaTuffaha:
Google and FCM removed

bahaaTuffaha:
My bad I shouldn't requests a pull request here

